PR #22, opened 2020-11-24
Initial work on a basic rate limit for spans

johnbley:
Also update batch period to 5s from 100ms.


karma.conf.js
```diff
@@ -92,8 +92,7 @@ module.exports = function (config) {
 
     // start these browsers
     // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-    browsers: ['Chrome', 'Firefox'],
-    //browsers: ['Chrome'], // useful to speed up edit+test cycle
+    browsers: ['Chrome'],
```
jtmalinowski, line 95: alright, first thing next week I'm going to stabilise the integration tests

src/main.js
```diff
@@ -122,8 +122,8 @@ if (!window.SplunkRum) {
     if (options.beaconUrl) {
       const completeUrl = options.beaconUrl + (options.rumAuth ? '?auth='+options.rumAuth : '');
       const batchSpanProcessor = new BatchSpanProcessor(new PatchedZipkinExporter(completeUrl), {
-        bufferTimeout: 100, //millis, tradeoff between batching and loss of spans by not sending before page close
-        bufferSize: 20, // spans, traceoff between batching and hitting sendBeacon invididual limits
+        bufferTimeout: 5000, //millis, tradeoff between batching and loss of spans by not sending before page close
```
jtmalinowski, line 125: should we discuss in the comment for this line, why the particular value was chosen?

src/zipkin.js
```diff
@@ -30,9 +32,24 @@ export function truncate(zspan) {
...
     this.beaconUrl = beaconUrl;
+    this.spanCounts = {};
+    const exporter = this;
+    setInterval(() => {
+      exporter.spanCounts = {};
+    }, SPAN_RATE_LIMIT_PERIOD);
+  }
+
+  filter(span) {
+    const component = span.attributes ? span.attributes.component || 'unknown' : 'unknown';
```
jtmalinowski, line 43: How about `(span.attributes && span.attributes.component) || 'unknown'`?

src/zipkin.js
```diff
@@ -30,9 +32,24 @@ export function truncate(zspan) {
 export class PatchedZipkinExporter {
   constructor(beaconUrl) {
     this.beaconUrl = beaconUrl;
+    this.spanCounts = {};
+    const exporter = this;
+    setInterval(() => {
+      exporter.spanCounts = {};
```
jtmalinowski, line 38: just `this.spanCounts = {};` should work fine

src/zipkin.js
johnbley: Lint didn't seem to like that.  Tried an alternate that doesn't repeat the constant at least.

src/main.js
johnbley: I'll make a note that it's still under experimentation...

PR #25, opened 2020-12-02
Initial webvitals integration

johnbley:


src/webvitals.js
```diff
@@ -0,0 +1,43 @@
...
+  const value = metric.value;
+  const span = tracer.startSpan('webvitals');
+  span.setAttribute(name, value);
+  span.end(span.startTime);
+}
+
+export function initWebVitals(provider) {
+  const tracer = provider.getTracer('webvitals');
+  // CLS is defined as being sent more than once, easier to just ensure that everything is sent just on the first occurence.
+  const reported = {};
```
jtmalinowski, line 33: you can make this var local to this file, because webvitals should be initiated at most once per page

src/webvitals.js
```diff
@@ -0,0 +1,43 @@
...
+
+function report(tracer, name, metric, reported) {
+  if (reported[name]) {
+    return;
+  }
+  reported[name] = true;
+  const value = metric.value;
+  const span = tracer.startSpan('webvitals');
+  span.setAttribute(name, value);
+  span.end(span.startTime);
```
jtmalinowski, line 27: have we confirmed this with Justin? or are we deciding now it's going to be sent like that

integration-tests/tests/webvitals/webvitals.spec.js
```diff
@@ -0,0 +1,55 @@
...
+      return;
+    }
+
+    const url = browser.globals.getUrl('/webvitals/webvitals.ejs');
+    await browser.url(url);
+    // fid won't happen unless there is interaction, so simulate a bit of that
+    await browser.click('#clicky');
+    await browser.waitForElementPresent('#p2');
+
+    await browser.execute(function() {
```
jtmalinowski, line 32: Can you please add a comment that this is a workaround for headless Chrome not emitting `visibilitychange`?

src/webvitals.js
johnbley: Excellent point.  I'll post on the team channel and seek some input.

integration-tests/tests/webvitals/webvitals.spec.js
johnbley: Good catch, will do.

src/webvitals.js
jtmalinowski: I was going to approve this, but now I'm not sure, I can't seem to find any conclusions in that Slack thread.

johnbley: I read it as "looks good enough" but we can wait another day or two for any more objections/discussion to pop up.

jtmalinowski (approved):

PR #26, opened 2020-12-03
Yet more docload testing

johnbley:
link.* fields for documentFetch and resourceFetch, resource fetch for 404 image.

### DIFF
--- a/integration-tests/tests/docload/docload.spec.js
+++ b/integration-tests/tests/docload/docload.spec.js
@@ -41,16 +41,27 @@ module.exports = {
 
     const url = browser.globals.getUrl('/docload/docload.ejs');
     await browser.url(url);
+
     const docFetch = await browser.globals.findSpan(span => span.name === 'documentFetch');
     const scriptFetch = await browser.globals.findSpan(span => span.name === 'resourceFetch' && span.tags['http.url'].includes('splunk-rum.js'));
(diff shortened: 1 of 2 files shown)


integration-tests/tests/docload/docload.spec.js
```diff
@@ -44,10 +44,13 @@ module.exports = {
     const docFetch = await browser.globals.findSpan(span => span.name === 'documentFetch');
     const scriptFetch = await browser.globals.findSpan(span => span.name === 'resourceFetch' && span.tags['http.url'].includes('splunk-rum.js'));
     const docLoad = await browser.globals.findSpan(span => span.name === 'documentLoad');
+
+    const brokenImgFetch = await browser.globals.findSpan(span => span.name === 'resourceFetch' && span.tags['http.url'].includes('/nosuchimage.jpg'));
```
jtmalinowski, line 48: not an issue, but for my own knowledge: wouldn't you say that if you have an  in a page, then its failure to fetch is a child span of the docload span, and consequently it's part of the same trace?

jtmalinowski (approved):

integration-tests/tests/docload/docload.spec.js
johnbley: Great call, these all should have the same traceId - I'll add those asserts.

PR #28, opened 2020-12-08
Integration tests cleanup

jtmalinowski:
[x] replaces waiting on spans-not-sent, with waiting on guard spans sent after the unintended spans
[x] lowers buffer timeout to send spans immediately
[ ] marks spec result in BrowserStack


jtmal-signalfx:
Needs to be reopened from an account supported by circle CI.

PR #29, opened 2020-12-08
Same site cookies

mhennoch:
- Depending if script loaded in iframe add SameSite attribute to cookie


integration-tests/tests/cookies/cookies.ejs
```diff
@@ -0,0 +1,14 @@
...
+<head>
+	<meta charset="UTF-8">
+	<title>Cookies</title>
+
+	<%- renderAgent({debug: true}) %>
+</head>
+<body>
+	<h1>COOKIES</h1>
+
+  <iframe src="/cookies/iframe.ejs"></iframe>
```
jtmalinowski, line 12: does it matter that the iframe has the same origin? I'd bet it does

src/main.js
```diff
@@ -68,16 +68,27 @@ if (!window.SplunkRum) {
...
-    let rumSessionId = undefined;
+    let rumSessionId = instanceId;
     const cookieSetter = function() {
       if (!document.cookie.includes(cookieName)) {
         const sessionId = generateId(128);
-        document.cookie = cookieName + '=' + sessionId + '; path=/; max-age=' + SessionTimeoutSeconds ;
+        let cookie = cookieName + '=' + sessionId + '; path=/; max-age=' + SessionTimeoutSeconds ;
+
+        if (isIframe()) {
+          cookie += ';SameSite=None; Secure';
```
jtmalinowski, line 81: How about we start building the decision doc in an .md file? Would be great to add explanations and reasoning behind all new decision together with the code to implement them.

src/utils.js
```diff
@@ -40,3 +40,11 @@ export function limitLen(s, cap) {
     return s;
   }
 }
+
+export function isIframe() {
```
jtmalinowski, line 44: I can see usefulness of exposing this and testing e2e across browsers, that said, I don't know what's the propensity of this to break, so you decide Marten, because you did the research.

src/main.js
mhennoch: Yep, good idea, I've thought about it also.

src/utils.js
mhennoch: Supposedly this is the safest and most widely supported way according to stackoverflow :P. 
Integration tests should cover this. Even if this breaks nothing would break for our users.

integration-tests/tests/cookies/cookies.ejs
mhennoch: It does but we don't have a good way of testing third party iframe. I could start another http server on another port or use some third party  external page somehow. As this is a corner case anyways the test only checks if correct cookie string is used in iframe vs non iframe.

johnbley: Having the capability to serve from a second port would be awesome for, e.g., testing CORS scenarios.  Let's throw that on the backlog but I think what you have here is fine for now.

src/main.js
johnbley: 👍

johnbley (approved):

PR #33, opened 2020-12-15
feat: foundation for profiling/benchmarking

jtmalinowski:


lib/server/index.js
```diff
@@ -0,0 +1,77 @@
+/*
```
johnbley, line 1: As I understand it from reading through this, the `./lib` directory is for code shared between performance-tests and integration-tests?  Perhaps `test-utils` which clearly marks this code as non-production, rather than the amibiguous `lib`?

johnbley (changes requested):

src/main.js
```diff
@@ -58,7 +58,7 @@ if (!window.SplunkRum) {
     if (!options.debug) {
       if (!options.beaconUrl) {
         throw new Error('SplunkRum.init( {beaconUrl: \'https://something\'} ) is required.');
-      } else if(!options.beaconUrl.startsWith('https') && !options.allowInsecureBeacon) {
+      } else if(options.beaconUrl.startsWith('http:') && !options.allowInsecureBeacon) {
```
johnbley, line 61: Without `allowInsecureBeacon` we shouldn't allow a relative `beaconUrl: '/beacon'` unless it resolves to an https url.

johnbley (changes requested):

src/main.js
jtmalinowski: as discussed in Slack, the `allowInsecureBeacon` part is removed from this PR

johnbley (approved):

PR #39, opened 2020-12-21
More xhr/fetch tests

johnbley:
- Rounding out coverage of events and attributes (WIP)


integration-tests/tests/spec-utils.js
```diff
@@ -0,0 +1,43 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+module.exports = {
+
+  timesMakeSense: async function(browser, eventsArr, startName, endName) {
```
jtmalinowski, line 19: maybe halpful: If this function was attached to `browser`, we could save typing thanks to one argument less, maybe something like `browser.timesMakeSense = (...args) => specUtils.timesMakeSense(browser, ...args)`

johnbley: Yes, much cleaner, thanks.

jtmalinowski (approved):

PR #40, opened 2021-01-04
Integration tests for user-interaction

jtmalinowski:


src/interaction.js
```diff
@@ -16,21 +16,31 @@ limitations under the License.
...
-  dragend: true,
-  drop: true,
-  ended: true,
-  pause: true,
-  play: true,
-  change: true,
-  mousedown: true,
-  mouseup: true,
-};
+const allowedEventTypes = [
```
mhennoch, line 19: An idea for future: maybe we should let user configure which of these they want to enable/disable. So if they are not interested in some events they can save bandwidth.

mhennoch (commented):
I think we should also test for causality. Eg. button click creates a new xhr/fetch and click event and xhr event have same traceId. This is probably where most of the bugs are.

src/interaction.js
jtmalinowski: Added to the tracker

src/interaction.js
```diff
@@ -16,21 +16,31 @@ limitations under the License.
...
+  // drap & drop
+  'dragend',
+  'drop',
+
+  // media
+  'ended',
+  'pause',
+  'play',
+
+  // keyboard
```
johnbley, line 40: Two things: 1) the style change looks OK, but won't array.includes change our runtime here from O(1) to O(n) (n=number of allowed events), and (bigger) I specifically don't have key events in here because I found that real-world apps (including some of our beta customers) trigger a storm of these on things like search fields (each key triggers a lookup in the search cache/pre-fetched suggestions, etc. -> each key is a span).  I like Marten's idea of exposing this through config but I think the default needs to keep keydown/keyup out.

johnbley (changes requested):

src/interaction.js
jtmalinowski: You're right, I'm dropping key events from defaults. With regards to array vs. set, iterating over an array of a dozen, maybe two, of items will add measurable overhead to an event handler. That said, object with boolean values makes more sense for configurability.

src/interaction.js
```diff
@@ -16,30 +16,43 @@ limitations under the License.
...
-  change: true,
-  mousedown: true,
-  mouseup: true,
 };
 
-
 export class SplunkUserInteractionPlugin extends UserInteractionPlugin {
+  constructor({adjustAutoInstrumentedEvents}) {
+    super();
+    this._autoInstrumentedEvents = adjustAutoInstrumentedEvents ?
```
mhennoch, line 44: Isn't `adjustAutoInstrumentedEvents` always true as { } is default?

README.md
```diff
@@ -67,6 +67,7 @@ Please read [INSTALLING.md](./docs/INSTALLING.md) for more advanced installation
 | debug | boolean | Turns on/off internal debug logging | No | false |
 | ignoreUrls | array | Applies for XHR,Fetch and Websocket URLs. URLs that partially match any regex in ignoreUrls will not be traced. In addition, URLs that are _exact matches_ of strings in ignoreUrls will also not be traced. | No | [] |
 | spanProcessor | SpanProcessor | Offers ability to alter/remove data in-browser.  See below for more details | No | (undefined) |
+| adjustAutoInstrumentedEvents | { [DOM Event Name]?: boolean } | Set keys to `false` to disable events handled by default. Set additional keys to true to auto-instrument `addEventListener` handlers. | No | Please check `window.SplunkRum.DEFAULT_AUTO_INSTRUMENTED_EVENTS` |
```
johnbley, line 70: This should be named `autoInstrumentedEvents`.

src/main.js
```diff
@@ -33,12 +33,20 @@ function browserSupported() {
...
+const OPTIONS_DEFAULTS = {
+  app: 'unknown-browser-app',
+  adjustAutoInstrumentedEvents: {},
+};
+
 if (!window.SplunkRum) {
   window.SplunkRum = {
-    inited: false
+    inited: false,
+    DEFAULT_AUTO_INSTRUMENTED_EVENTS,
```
johnbley, line 44: I like the idea of exposing this so the defaults can be tweaked easily.

johnbley (changes requested):

johnbley (approved):

mhennoch (approved):

PR #41, opened 2021-01-05
GH actions CI run linkage with browserstack

jtmalinowski:


johnbley (approved):
Seems like a good idea to me.

PR #42, opened 2021-01-07
Use environment instead of deployment.environment per team discussion

johnbley:


test/init.test.js
```diff
@@ -141,7 +141,7 @@ describe('creating spans is possible', () => {
       assert.ok(!!span.attributes['location.href']);
       assert.ok(!!span.attributes['splunk.scriptInstance']);
       assert.strictEqual(span.attributes['app'], 'my-app');
-      assert.strictEqual(span.attributes['deployment.environment'], 'my-env');
+      assert.strictEqual(span.attributes['environment'], 'my-env');
```
mhennoch, line 144: Maybe add a simple integration test for this also?

mhennoch (approved):

test/init.test.js
johnbley: I'll go one further: I created a task to make an integration test for all the `init()` config options, which will cover this.

PR #44, opened 2021-01-11
Add a bit of documentation around our use of Server-Timing.

johnbley:


docs/ServerTraceContext.md
```diff
@@ -0,0 +1,32 @@
+# Server Trace Context
+
+## Rationale
+
+`splunk-rum.js` captures the server trace context from a `Server-Timing` header in the `traceparent` format.  Our
+server-side instrumentation generates this automatically.  This document describes how to recreate this if you want to
```
mhennoch, line 6: Our as in signalfx agents? Maybe we should mention that no support in otel as of yet.  We could also list the language agents that do support it if not all signalfx agents don't.

docs/ServerTraceContext.md
```diff
@@ -0,0 +1,32 @@
...
+
+Please familiarize yourself with these specifications:
+- <https://www.w3.org/TR/server-timing/>
+- <https://www.w3.org/TR/trace-context/#traceparent-header>
+
+We expect a Server-Timing entry with the name `traceparent` where the `desc` field holds the
+corresponding value:
+
+```http
+Server-Timing: traceparent;desc="00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00"
```
mhennoch, line 19: Doesn't splunk-rum expect sampled flag to be 01 (https://github.com/signalfx/splunk-browser-otel-js/blob/main/src/servertiming.js#L27)?

johnbley: Excellent catch!

docs/ServerTraceContext.md
johnbley: Yes, I'll note that it's just the SignalFx agents rather than trying to keep a detailed list here.

mhennoch (approved):

PR #46, opened 2021-01-13
Init options tests

johnbley:


integration-tests/tests/init/init.spec.js
```diff
@@ -16,8 +16,33 @@ limitations under the License.
...
+    // environment still set (not cleared by setGlobalAttributes call)
+    await browser.assert.strictEqual(atts2.tags['environment'], 'custom-environment');
+    await browser.assert.strictEqual(atts2.tags['key1'], 'newvalue1');
+    await browser.assert.strictEqual(atts2.tags['key2'], undefined);
+  },
+  'captureErrors controls error capture': async function(browser) {
+    browser.globals.clearReceivedSpans();
+    await browser.url(browser.globals.getUrl('/init/captureErrors.ejs'));
+    const lackOfError = await browser.globals.findSpan(span => span.tags.component === 'error');
+    await browser.assert.not.ok(lackOfError);
```
jtmalinowski, line 46: We can make this a bit more stable by sending a guard span after this one, awaiting the guard span, and finally checking the list of received spans. Example: https://github.com/signalfx/splunk-browser-otel-js/blob/main/integration-tests/tests/websocket/websockets.spec.js#L52

johnbley: Great idea, I'll put that in later today.

integration-tests/tests/init/captureErrors.ejs
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<title>Test init</title>
+
+	<%- renderAgent({captureErrors: false}) %>
+</head>
+<body>
+	<p>Attributes</p>
```
mhennoch, line 10: I like to use different html content so differentiating tests on browserstack video is easier.

mhennoch (approved):

PR #49, opened 2021-01-19
Don't throw SyntaxError on IE11

mhennoch:
**Short version**

1. transpile our .js code so no syntaxerrors are thrown.
2. throw early with meaningful error message if browser does not support needed features.

**Some info and notes for future self.**

In order to work in IE11 fully we would need to:
1. polyfill otel code
2. polyfill sendBeacon and possibly also something that works on page onload eg. image.
3. transpile our .js code

2 and 3 are quite easy (Just transpiling our .js code only adds about 2-4kb.
) but 1. will make our bundle size from 76kb to 124kb(unzipped) give or take. Gzipped 20 vs 40kb. Also a lot of the tests seem to fail but fundamentally getting pageload spans and xhr should(or can be made to) work.  This polyfilling was done via rollup babel plugin and corejs using "useBuiltIns": "usage" option.

Some of the things that would need to be polyfilled:
```
  es.function.name { "ie":"11" }
  es.object.keys { "ie":"11" }
  es.symbol { "ie":"11" }
  es.symbol.description { "ie":"11" }
  es.array.iterator { "ie":"11" }
  es.map { "ie":"11" }
  es.object.to-string { "ie":"11" }
  es.string.iterator { "ie":"11" }
  web.dom-collections.iterator { "ie":"11" }
  es.global-this { "ie":"11" }
  ...
```
We could also provide separate bundle for IE11 or a separate .js file with the polyfills. Another issue to think about is that if someone heavily relies on otel API for manual instrumentation(tracer.withSpan) etc. and otel fails to load the app will most likely be broken. When using npm for adding splunkrum a lot of this can be mitigated.

Some precautions when using API:
- Use try catch or if(SplunkRum) {} when instrumenting manually.
- Bundle noop tracer with the app as a fallback ( we could provide an easy way to get it maybe)


src/oldBrowserSupportCheck.js
```diff
@@ -0,0 +1,20 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+if( typeof Symbol !== 'function') {
+  window.SplunkRum = { init: function(){} };
```
johnbley, line 18: Perhaps add no-op functions for `error` and `setGlobalAttributes` as well?

johnbley (changes requested):

src/oldBrowserSupportCheck.js
mhennoch: Good catch

johnbley (approved):

rollup.config.js
```diff
@@ -18,21 +19,25 @@ export default {
...
     commonjs({
       include: /node_modules/,
     }),
     process.env.DEBUG_BUILD ? {} :
-    terser({
-      output: { comments: false, },
-    }),
+      terser({
+        output: { comments: false, },
+      }),
```
jtmalinowski, line 41: can we keep the indentation sane please? :)

src/oldBrowserSupportCheck.js
```diff
@@ -0,0 +1,25 @@
+/*
+Copyright 2020 Splunk Inc.
```
jtmalinowski, line 2: If you rebase to master, remove the header and regenerate it, you should start getting the current year

babel.config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "presets": [
+    [
+      "@babel/env",
+      {
+        "targets": {
```
jtmalinowski, line 6: Since you have started this file, I'd say that it should serve as a true source of which browsers are supported, and I'd add info about other browsers here as well, even though it will likely not affect babel's behaviour.

PR #52, opened 2021-01-21
Inactivity session timeout

johnbley:


src/session.js
```diff
@@ -0,0 +1,134 @@
...
+    renewCookieTimeout(sessionState);
+  }
+  recentActivity = false;
+}
+
+export function initSessionTracking(instanceId) {
+  rumSessionId = instanceId;
+  recentActivity = true; // document loaded implies activity
+  updateSessionStatus();
+  setInterval(updateSessionStatus, PeriodicCheckSeconds * 1000);
```
jtmalinowski, line 125: Looks like a ducktape for _.throttle (<https://lodash.com/docs/4.17.15#throttle>). For our case I would add `{leading: true, trailing: true}`, which should mirror activity callbacks more closely.

src/session.js
```diff
@@ -0,0 +1,134 @@
...
+
+const MaxSessionAgeMillis = 4 * 60 * 60 * 1000;
+const InactivityTimeoutSeconds = 15 * 60;
+const PeriodicCheckSeconds = 60;
+export const COOKIE_NAME = '_splunk_rum_sid';
+
+let rumSessionId = generateId(64); // will be overwritten in init() with scriptInstance; only here as extreme fallback
+let recentActivity = false;
+function markActivity() {
+  recentActivity = true;
```
jtmalinowski, line 44: `lastActivityTime = performance.now()` would be more readable, otherwise you have to guess when was the last time it was cleared, which could be difficult if a bug had something to do with inconsistency of `setInverval`

test/session.test.js
```diff
@@ -0,0 +1,57 @@
...
+    initSessionTracking('1234');
+    const firstSessionId = getRumSessionId();
+    assert.strictEqual(firstSessionId.length, 32);
+    // no marked activity, should keep same state
+    updateSessionStatus();
+    assert.strictEqual(firstSessionId, getRumSessionId());
+    // set cookie to expire in 2 seconds, mark activity, and then updateSessionStatus.  
+    // Wait 4 seconds and cookie should still be there (having been renewed)
+    const cookieValue = encodeURIComponent(JSON.stringify({id:firstSessionId, startTime: new Date().getTime()}));
+    document.cookie = COOKIE_NAME + '=' + cookieValue + '; path=/; max-age=' + 4;
```
jtmalinowski, line 34: But you're not setting 2 seconds here, you're setting 4, no? If I understand correctly this value should definitely be lower than what you're putting in `setTimeout`.

jtmalinowski (changes requested):

test/session.test.js
johnbley: Whoops!

src/session.js
johnbley: I'd just as soon not pull in another library unless the complexity of this thing grows substantially bigger.  Also, it's important for the behavior of this that the first `updateSessionStatus` happen before `init` returns, rather than on next tick...

src/session.js
johnbley: Not sure if you meant it this way, but I gather a concern here is the idea that we could mark recent activity, the tab is backgrounded, and an hour later the tab is foregrounded with `updateSessionStatus` running only then (and seeing the "old" `recentActivity`).  Assuming the browser does in fact expire the session for us in the interim, the first thing the periodic task will do is notice the expired cookie and start a new one -> no false continuation of the stale session.  That said, it would be cleaner/timelier to make sure that `updateSessionStatus` also gets called on foregrounding, so I'll add that.

src/session.js
```diff
@@ -0,0 +1,134 @@
...
+        id: 'sessionIdAsHex',
+        startTime: startTimeAsNewDate_getTime
+    }
+    Future work can add more fields though note that the fact that the value doesn't change 
+    once created makes this very robust when used in multiple tabs/windows - tabs don't compete/
+    race to do anything but set the max-age.
+*/
+
+
+const MaxSessionAgeMillis = 4 * 60 * 60 * 1000;
```
mhennoch, line 36: These 3 constants should also follow XXX_XXX style.

mhennoch (commented):
Document our session definition in the README?

src/session.js
johnbley: Tried that and got ` Parsing error: Identifier directly after number` from lint.  Might be related to the ecmaVersion defined in the lint config which is sorta obviated by the introduction of babel; not interested in going down that rabbit hole at this time.

PR #60, opened 2021-02-02
Capture resources loaded after page load event

mhennoch:
**Initial version of "ResourceObserver"(I am open for better name)**

With default conf creates new span for every image and script loaded after page load.
Can be configured to track more entry types but this is not covered via tests at the moment so the conf option is not documented.
Urls can be ignored.

**Some thoughts for future**
Maybe needs some extra work to 100% avoid duplicating spans. Could do `performance.getEntriesByType('resource')` on load and check if entry with some name already exists for example.


integration-tests/tests/resource-observer/resources.ejs
```diff
@@ -0,0 +1,49 @@
...
+	<title>Resource observer test</title>
+  <%- renderAgent() %>
+</head>
+<body>
+  <h1>Resource observer test</h1>
+
+  <button id="load-img">Load new Image</button>
+  <button id="load-script">Load new script</button>
+
+  <img id="img-1" src="/integration-tests/assets/splunk-black-white.png" alt="black">
```
jtmalinowski, line 14: Can you move https://github.com/signalfx/splunk-browser-otel-js/blob/main/performance-tests/devServer/randomImageProvider.js to test-utils, and mount in the integration tests server instead of these static images? It might get a bit more tricky if we wanted to assert content length, but could be solved with hard-coded seed.

integration-tests/tests/resource-observer/resources.ejs
```diff
@@ -0,0 +1,49 @@
...
+
+  <script>
+
+    document.getElementById('load-img').addEventListener('click', function() {
+      // random third party img
+      var img = new Image();
+      img.src = 'https://placeimg.com/320/240/any';
```
jtmalinowski, line 21: It doesn't seem like a good idea to rely on an external, I'm wondering what the point of doing this is?

integration-tests/tests/resource-observer/resources.ejs
```diff
@@ -0,0 +1,49 @@
...
+
+    document.getElementById('load-script').addEventListener('click', function() {
+      loadScript();
+    });
+
+    window.addEventListener("load", function() {
+      setTimeout(function() {
+        loadImg();
+        loadScript();
+      }, 500)
```
jtmalinowski, line 44: What does this achieve that
```
document.getElementById('load-script').addEventListener('click', loadScript);
```
doesn't? Assuming you extend it to handle images.

integration-tests/tests/resource-observer/resource-obs.spec.js
```diff
@@ -0,0 +1,50 @@
...
+module.exports = {
+  'should report resource loads happening after page load': async function(browser) {
+    browser.globals.clearReceivedSpans();
+    await browser.url(browser.globals.getUrl('/resource-observer/resources.ejs'));
+
+    const imageSpan = await browser.globals.findSpan(
+      span => span.tags['http.url'] && span.tags['http.url'].endsWith('splunk-white-black.png')
+    );
+
+    await browser.assert.ok(imageSpan, 'Image span not found.');
```
jtmalinowski, line 26: AFAIK the string which is the second parameter is not a failure message, but rather a name or a description of the assertion.

integration-tests/tests/resource-observer/resource-obs.spec.js
```diff
@@ -0,0 +1,50 @@
...
+limitations under the License.
+*/
+
+module.exports = {
+  'should report resource loads happening after page load': async function(browser) {
+    browser.globals.clearReceivedSpans();
+    await browser.url(browser.globals.getUrl('/resource-observer/resources.ejs'));
+
+    const imageSpan = await browser.globals.findSpan(
+      span => span.tags['http.url'] && span.tags['http.url'].endsWith('splunk-white-black.png')
```
jtmalinowski, line 23: Seems like this image is also in the page source? How are you ensuring the span is from after the page is loaded?

integration-tests/tests/resource-observer/resource-obs.spec.js
```diff
@@ -0,0 +1,50 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+module.exports = {
+  'should report resource loads happening after page load': async function(browser) {
+    browser.globals.clearReceivedSpans();
+    await browser.url(browser.globals.getUrl('/resource-observer/resources.ejs'));
+
+    const imageSpan = await browser.globals.findSpan(
```
jtmalinowski, line 22: You should send a guard span at the point when the page is loaded and wait for it here. You can grep the project by "guard", and just copy paste the implementation.

src/resource-observer.js
```diff
@@ -0,0 +1,95 @@
...
+  InstrumentationBase,
+} from '@opentelemetry/instrumentation';
+
+import { version } from '../package.json';
+import { hrTime, isUrlIgnored } from '@opentelemetry/core';
+import { addSpanNetworkEvents } from '@opentelemetry/web';
+import { HttpAttribute } from '@opentelemetry/semantic-conventions';
+
+const MODULE_NAME = 'splunk-resource-obs';
+const defaultAllowedInitiatorTypes = ['img', 'script']; //other, css, link
```
jtmalinowski, line 27: yes, what about link?

jtmalinowski (changes requested):

jtmalinowski:
How about PostDocLoadResourceObserver?

integration-tests/tests/resource-observer/resources.ejs
mhennoch: I use test files for manual testing. This is a nice endpoint for it been using it for years. 
Random sizes and third party.

integration-tests/tests/resource-observer/resources.ejs
mhennoch: I don't have to click

integration-tests/tests/resource-observer/resource-obs.spec.js
mhennoch: I actually do use it for the ignore test already but good idea, I'll use it here also.

integration-tests/tests/resource-observer/resources.ejs
mhennoch: What's wrong with static images? Every test that needs an image can just use this one. Pretty much need one of each asset (css, img, .js) for all of the tests. Actually this image tag is even unneeded here I'll just remove it.

integration-tests/tests/resource-observer/resources.ejs
johnbley: I agree, let's not depend on third parties for basic testing.

integration-tests/tests/resource-observer/resource-obs.spec.js
jtmalinowski: Gotcha, in that case I would change the names so they are more distinctive.

integration-tests/tests/resource-observer/resources.ejs
jtmalinowski: Doesn't seem useful to me, and adds guaranteed delay.

integration-tests/tests/resource-observer/resource-obs.spec.js
jtmalinowski: I don't see any guard spans in the ignore url test. What I mean is https://github.com/signalfx/splunk-browser-otel-js/blob/main/integration-tests/tests/fetch/fetch.spec.js#L46

jtmalinowski: Right, my bad.

integration-tests/tests/resource-observer/resources.ejs
mhennoch: removed, was just for testing anyways

integration-tests/tests/resource-observer/resource-obs.spec.js
```diff
@@ -0,0 +1,50 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+module.exports = {
+  'should report resource loads happening after page load': async function(browser) {
+    browser.globals.clearReceivedSpans();
```
johnbley, line 19: Perhaps it's here and I'm not seeing it, but can we get a test that ensures that we don't now double-report resources from the docload sequence?  I.e., there should be only 1 span for an image that is present in the original HTML, not 2.

mhennoch: It depends if the image is cached or not. If it is not cached then there should be a span if cached and no network request done then no span. Added two new tests to test this cache/no-cache situation.

integration-tests/tests/resource-observer/resources.ejs
mhennoch: As discussed in slack, lowered it.

src/resource-observer.js
mhennoch: I think other entry types will be just noise right now, some third party stuff etc. Most of this is loaded during page load for most sites. People can turn them on if they know what they are doing.

integration-tests/tests/resource-observer/resource-obs.spec.js
johnbley: I just meant that by having a separate observer we run the risk of reporting two spans for the same original performance event, regardless of cache/no-cache.  Adding cache/no-cache tests is great too, but I was asking for something simpler.

integration-tests/tests/resource-observer/resources-ignored.ejs
```diff
@@ -0,0 +1,51 @@
+<!DOCTYPE html>
```
jtmalinowski, line 1: can we merge the view templates into one, and control using query params?

jtmalinowski (changes requested):

integration-tests/tests/resource-observer/resources-ignored.ejs
mhennoch: We can but I prefer extra template files.

integration-tests/tests/resource-observer/resource-obs.spec.js
mhennoch: Added asserts checking that there's only one span for resources loaded during page load.

integration-tests/tests/resource-observer/resources-ignored.ejs
mhennoch: At some point it might make sense but I don't think it does here

integration-tests/tests/resource-observer/resources.ejs
```diff
@@ -0,0 +1,43 @@
...
+<body>
+  <h1>Resource observer test</h1>
+  <img src="/integration-tests/assets/no-cache.png" alt="no-cache">
+  
+  <script>
+    function loadScript() {
+      var p = new Promise(function(resolve, _) {
+        var s = document.createElement( 'script' );
+        s.setAttribute( 'src', '/integration-tests/assets/test.js' );
+        s.onload = resolve();
```
jtmalinowski, line 17: should be `s.onload = resolve;`

src/postDocLoadResourceObserver.js
```diff
@@ -0,0 +1,92 @@
...
+      span.end();
+    }
+  }
+
+  enable() {
+    if (window.PerformanceObserver) {
+      if (window.document.readyState === 'complete') {
+        this._startObserver();
+      } else {
+        window.addEventListener('load', () => {
```
jtmalinowski, line 80: I think this is the reason why you need `setTimeout(..., 5)` in `load` handler in tests, especially that you set `buffered: false` when starting the observer. There are competing `load` handlers. Looks to me like this increases flakiness overall, and needs either: 1. checking `readyState === 'complete'` within observer callback, or even better looking at the timing of each entry to determine if it happened before of after load. Could it be that the observer may send performance entries from before and after `load` in one callback?

jtmalinowski (changes requested):

src/postDocLoadResourceObserver.js
```diff
@@ -0,0 +1,92 @@
...
+import { HttpAttribute } from '@opentelemetry/semantic-conventions';
+
+const MODULE_NAME = 'splunk-post-doc-load-resource';
+const defaultAllowedInitiatorTypes = ['img', 'script']; //other, css, link
+export class PostDocLoadResourceObserver extends InstrumentationBase {
+  
+  constructor(config = {}) {
+    super(MODULE_NAME, version, Object.assign({}, {allowedInitiatorTypes: defaultAllowedInitiatorTypes}, config));
+    this.observer = undefined;
+    this.afterInit = false;
```
jtmalinowski, line 33: This can now be removed.

src/postDocLoadResourceObserver.js
mhennoch: Changed it to readyState checking in observer callback

jtmalinowski (approved):

PR #65, opened 2021-02-15
Fix fetch url ignoring. 

mhennoch:
1. Fetch url ignoring broken at the moment.  Fetch Instrumentation constructor takes config not provider and config.
2. Test for it was also broken because fetch spans are ended 300ms after fetch end in order to make sure timings are present when ending span.  So guard-span arrived before fetch spans and test passed although it was broken.


jtmalinowski (approved):
Please decrease the 1000ms to 310ms. I'm pretty sure that spans will reach the backend in the same order that they are ended so we don't need any buffers.

PR #66, opened 2021-02-15
Add delta threshold to resource timing filtering

mhennoch:
Add some margin so tests are not flaky. See fork commit for info. I've seen this happen with latest chrome both headless and normal. Probably can happen with all of them


jtmalinowski (changes requested):
Can you please:
- add in the comment which browser and version this fix is for
- squash the fork commits
- rename the distro commit so it matches the fork commits content?

jtmalinowski (approved):

PR #67, opened 2021-02-16
Safari10 support

mhennoch:
1. As upstream is updated this now works, issue was that getEntriesByType being missing was not accounted for.
2. Fix bug with XHR headers in safari10, case sensitivity issue.
3. add window.PerformanceObserver otherwise the whole script fails.
4. Added safari10.1 target to browserstack, tagged some basic tests with safari10 tag.
5. Things are still quite flaky test wise, otherwise in real world it seems to work fineish.
6. Added XHR fallback to sendBeacon, ddog uses almost the same fallback.


integration-tests/utils/globals.js
```diff
@@ -52,7 +52,13 @@ module.exports = {
     browser.globals.receivedSpans = spans;
     browser.globals.rumVersion = require('../../package.json').version;
     browser.globals.clearReceivedSpans = () => { spans.length = 0; };
-    browser.globals.findSpan = (testFn, timeout = 0) => findSpan(spans, testFn, timeout);
+    let defaultTimeout = 0;
+    const {browserName, browser_version} = browser.options.desiredCapabilities;
+    if (browserName === 'Safari' && browser_version === '10.1') {
+      console.log('Setting default timeout for safari');
```
t2t2, line 58: turn into comment / note on why?

src/longtask.js
```diff
@@ -79,7 +79,7 @@ export class SplunkLongTaskInstrumentation extends InstrumentationBase {
 
   _isSupported() {
     // note: PerformanceObserver.supportedEntryTypes has better browser support than LongTask
-    const supportedEntryTypes = PerformanceObserver && PerformanceObserver.supportedEntryTypes;
+    const supportedEntryTypes = window.PerformanceObserver && window.PerformanceObserver.supportedEntryTypes;
```
t2t2, line 82: second `window` is redundant as first confirms it's available as global variable
```suggestion
    const supportedEntryTypes = window.PerformanceObserver && PerformanceObserver.supportedEntryTypes;
```

t2t2 (approved):

integration-tests/tests/docload/docload.spec.js
```diff
@@ -48,16 +52,14 @@ module.exports = {
     await browser.timesMakeSense(docFetch.annotations, 'requestStart', 'responseStart');
     await browser.timesMakeSense(docFetch.annotations, 'responseStart', 'responseEnd');
     await browser.timesMakeSense(docFetch.annotations, 'fetchStart', 'responseEnd');
-    if (browser.options.desiredCapabilities.browserName !== 'Safari') {
+    if (browserName !== 'Safari') {
```
jtmalinowski, line 55: Can we start extracting browser conditions to helpers? Or at least compare after lower-casing browser names?

integration-tests/tests/docload/docload.spec.js
```diff
@@ -15,31 +15,35 @@ limitations under the License.
 */
 
 module.exports = {
-
+  '@tags': ['safari-10.1'],
```
jtmalinowski, line 18: Does it do anything if you have explicit conditions for Safari everywhere?

integration-tests/tests/docload/docload.spec.js
jtmalinowski: On second thought, this is repeated so many times that it should be extracted to a helper.

jtmalinowski (changes requested):

integration-tests/tests/docload/docload.spec.js
mhennoch: Some tests won't ever work for safari 10 eg lontask, webvitals etc. So idea is to tag everything that should theoretically work with this tag so we can easily run them

jtmalinowski: Right, but you haven't modified any test runner methods, and I don't see the tag in use anywhere.

jtmalinowski (approved):

PR #68, opened 2021-02-17
Fix flaky test

mhennoch:


integration-tests/tests/resource-observer/resources-ignored.ejs
```diff
@@ -31,7 +31,7 @@
     function loadImg() {
       var p = new Promise(function(resolve, _) {
         var img = document.createElement('img');
-        img.src = '/integration-tests/assets/splunk-black.png?t=100';
+        img.src = '/integration-tests/assets/splunk-black.png';
```
jtmalinowski, line 34: can you explain why?

jtmalinowski (approved):

integration-tests/tests/resource-observer/resource-obs.spec.js
```diff
@@ -57,8 +57,12 @@ module.exports = {
     
     await browser.globals.findSpan(span => span.name === 'guard-span');
     const url = browser.globals.getUrl('/integration-tests/assets/', []);
-    await browser.assert.not.ok(browser.globals.getReceivedSpans().find(span => span.tags['http.url'] === url + 'test.js'));
-    const imgSpan = browser.globals.getReceivedSpans().find(span => span.tags['http.url'] === url + 'splunk-black.png');
+    await browser.assert.not.ok(browser.globals.getReceivedSpans().find(
+      span => span.tags['http.url'] === url + 'test.js' && span.tags['component'] === 'splunk-post-doc-load-resource'
```
johnbley, line 61: Was this failing because we double-reported the resource (once from docload and once from post-doc-load)?

mhennoch: No, sometimes they just ended up under page load for some reason. 
Adding a small timeout to resource load fixes it also. But a these tests seem a bit too flaky for various reasons but no 
double loads fortunately. But I am considering disabling this plugin by default for now or revisiting it before GA.

PR #76, opened 2021-03-01
Don't run some doc load tests on safari10

mhennoch:


jtmalinowski (changes requested):
Looks like it's missing a piece of code which controls which specs are run in each browser.

mhennoch:
As discussed in slack it makes more sense to blacklist test rather than whitelist.  But this is a larger piece of work.
This is just a quickfix for now.

mhennoch:
As discussed in slack it makes more sense to blacklist test rather than whitelist.  But this is a larger piece of work.
This is just a quickfix for now.

jtmalinowski (approved):

PR #96, opened 2021-03-24
Fix babel not transpiling our files

t2t2:
Noticed while manually testing in IE that there was an odd syntax error from for ... of statements instead of the error from `throwIfUnsupportedBrowser.js` as expected. Then noticed that only our files (everything in `src/` folders) weren't being touched by babel and were in their raw form in built files (class syntax and all). Removing the filter so absolutely all files get babel treatment

### DIFF
--- a/rollup.shared.js
+++ b/rollup.shared.js
@@ -6,7 +6,6 @@ const commonjs = require('@rollup/plugin-commonjs');
 
 module.exports.babelPlugin = babel({ 
   babelHelpers: 'runtime',
-  include: /node_modules/,
 });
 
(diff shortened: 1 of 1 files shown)


t2t2:
can also spot them in https://cdn.signalfx.com/o11y-gdi-rum/v0.3.1/splunk-otel-web.js if you ctrl/cmd+f for `class`, or `for(const`

jtmalinowski (approved):
I'm not sure class is a good example, because it might not be transpired anyway, looking at out target browsers, but this change makes sense.

PR #99, opened 2021-03-29
Allow redacting PII before exporting spans

jtmalinowski:


README.md
```diff
@@ -125,6 +126,28 @@ method. Using it will overwrite specified properties and leave others unchanged.
 Any spans reported from this point on will have your new attributes set.  
 You can pass `{}` or `undefined` to clear your global attributes.
 
+### Redacting PII
+In certain situations, metadata collected by our instrumentation may include PII.
+We'd advise that you review 2 cases in particular:
+- any network operation, where a secret piece of information might be present in the URL (e.g. an authentication token)
```
johnbley, line 132: "Please note that we do not capture or report any data from the payload of the request (i.e. the POST body), apart from its size"

src/SplunkExporter.js
```diff
@@ -0,0 +1,119 @@
...
+const SPAN_RATE_LIMIT_PERIOD = 30000; // millis, sweep to clear out span counts
+const MAX_SPANS_PER_PERIOD_PER_COMPONENT = 100;
+const SERVICE_NAME = 'browser';
+
+const SPAN_TO_ATTRIBUTES = (span) => span.attributes;
+const XHR_FACTORY = () => new XMLHttpRequest();
+const BEACON_SENDER = navigator.sendBeacon ? (url, data) => navigator.sendBeacon(url, data) : null;
+
+/**
+ * SplunkExporter is compatible with Zipkin V2 with some.
```
johnbley, line 36: ... with some what?  :-)

johnbley (approved):
Looks good; few minor comments on documentation.

src/index.js
```diff
@@ -51,12 +56,22 @@ const INSTRUMENTATIONS = [
...
 const NOOP = () => {};
 
+function buildExporter(options) {
+  const completeUrl = options.beaconUrl + (options.rumAuth ? '?auth='+options.rumAuth : '');
+  return options.exporter._factory({
+    beaconUrl: completeUrl,
+    onAttributesSerializing: options.exporter.onAttributesSerializing,
+  });
+}
+
```
t2t2, line 66: Do we need it to be a factory? Would simplify it down to

```js
return new SplunkExporter({
    beaconUrl: completeUrl,
    ...options.exporter
})
// or I'd even rather
return new SplunkExporter(completeUrl, options.exporter)
```

src/SplunkExporter.js
```diff
@@ -0,0 +1,119 @@
...
+import {ExportResultCode} from '@opentelemetry/core';
+import {limitLen} from './utils';
+import { Resource } from '@opentelemetry/resources';
+
+const MAX_VALUE_LIMIT = 4096;
+const SPAN_RATE_LIMIT_PERIOD = 30000; // millis, sweep to clear out span counts
+const MAX_SPANS_PER_PERIOD_PER_COMPONENT = 100;
+const SERVICE_NAME = 'browser';
+
+const SPAN_TO_ATTRIBUTES = (span) => span.attributes;
```
t2t2, line 31: Idk about this `(span) => attributes` input-output mismatch

1. Either missing out on chance to also apply this to other data on spans (exceptions _should_ be an event on some span, but that might be future), or to filter out spans (eg. return null)
2. Or how about `(attributes, span) => attributes`, since most will need `span.attributes` anyway and then can `(attrs) => ({...attrs, whatever})` for most cases, or use span from 2nd argument if they need it for some reason

README.md
```diff
@@ -125,6 +126,28 @@ method. Using it will overwrite specified properties and leave others unchanged.
...
+  ...otherOptions,
+  exporter: {
+    onAttributesSerializing: (span) => ({
+      ...span.attributes,
+      'http.url': /secret\=/.test(span.attributes['http.url']) ? '[redacted]' : span.attributes['http.url'],
+    }),
+  },
+});
+```
+Please note that this method (spread operator) carries a small risk of dropping non-enumerable properties should anyone define them on `span.attributes`.
```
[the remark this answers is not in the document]
jtmalinowski: I'll handle this on our end

src/SplunkExporter.js
```diff
@@ -0,0 +1,119 @@
...
+const XHR_FACTORY = () => new XMLHttpRequest();
+const BEACON_SENDER = navigator.sendBeacon ? (url, data) => navigator.sendBeacon(url, data) : null;
```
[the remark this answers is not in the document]
jtmalinowski: because they have different interfaces, but on second thought it'll maybe be a bit more convenient

src/index.js
jtmalinowski: It definitely should be configurable one way or another. a) feature parity with Nodejs bistro b) less coupling. If you have ideas for a better config format, let me know.

src/SplunkExporter.js
jtmalinowski: good point

t2t2 (approved):

PR #114, opened 2021-04-06
fix: don't wait for browser to end

jtmalinowski:
This causes nightwatch to get stuck if there's a failure in a spec at the end
of which we wait for the browser to end.


t2t2 (approved):

mhennoch (approved):
🙏

PR #126, opened 2021-04-15
Chore/type tests

jtmalinowski:


src/index.js
```diff
@@ -71,12 +76,14 @@ const SplunkRum = {
...
   init: function (options = {}) {
+    diag.setLogger(new DiagConsoleLogger(), options.debug ? DiagLogLevel.DEBUG : DiagLogLevel.ERROR);
+
     options = Object.assign({}, OPTIONS_DEFAULTS, options, {
       exporter: Object.assign({}, OPTIONS_DEFAULTS.exporter, options.exporter),
     });
 
     if (this.inited) {
-      console.log('SplunkRum already init()ed.');
+      diag.warn('SplunkRum already init()ed.');
```
t2t2, line 86: should this be error to match previous behavior of it being output to console regardless of options.debug?

jtmalinowski: Right, that makes sense.

mhennoch (approved):

PR #151, opened 2021-05-05
feat: allow reading and watching global attributes

jtmalinowski:


src/index.ts
```diff
@@ -308,7 +345,19 @@ const SplunkRum: SplunkOtelWebType = {
     }
 
     _errorInstrumentation.report('SplunkRum.error', args);
-  }
+  },
+
+  _experimental_addEventListener(name: SplunkOtelWebEventType, callback: (event: never) => void): void {
```
t2t2, line 350: What's the point of having internal event target methods also have experimental prefix when it's publicly only exposed here? Could get rid of the custom `NativeEventTarget`

src/EventTarget.ts
```diff
@@ -0,0 +1,68 @@
...
+  _experimental_addEventListener(name: SplunkOtelWebEventType.GlobalAttributesChanged, callback: (event: GlobalAttributesChangedEvent) => void);
+
+  _experimental_removeEventListener(name: SplunkOtelWebEventType.SessionChanged, callback: (event: SessionIdChangedEvent) => void);
+  _experimental_removeEventListener(name: SplunkOtelWebEventType.GlobalAttributesChanged, callback: (event: GlobalAttributesChangedEvent) => void);
+}
+
+export class NativeEventTarget implements SplunkOtelWebEventTarget {
+  // we're using DOM API to save on bytes a little bit
+  // another benefit is
+  private readonly target = new EventTarget;
```
t2t2, line 55: Bad compatibility in IE/safari <14 https://developer.mozilla.org/en-US/docs/Web/API/EventTarget#browser_compatibility

src/EventTarget.ts
```diff
@@ -0,0 +1,68 @@
...
+  _experimental_addEventListener(name: SplunkOtelWebEventType.SessionChanged, callback: (event: SessionIdChangedEvent) => void);
+  _experimental_addEventListener(name: SplunkOtelWebEventType.GlobalAttributesChanged, callback: (event: GlobalAttributesChangedEvent) => void);
+
+  _experimental_removeEventListener(name: SplunkOtelWebEventType.SessionChanged, callback: (event: SessionIdChangedEvent) => void);
+  _experimental_removeEventListener(name: SplunkOtelWebEventType.GlobalAttributesChanged, callback: (event: GlobalAttributesChangedEvent) => void);
+}
+
+export class NativeEventTarget implements SplunkOtelWebEventTarget {
+  // we're using DOM API to save on bytes a little bit
+  // another benefit is
```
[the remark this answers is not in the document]
jtmalinowski: fixed

src/index.ts
jtmalinowski: It's just convenient to reuse `SplunkOtelWebEventTarget` interface internally and externally.

src/EventTarget.ts
jtmalinowski: replaced with `HTMLDivElement`, good catch!

mhennoch (approved):

src/EventTarget.ts
```diff
@@ -41,28 +41,21 @@ export class GlobalAttributesChangedEvent extends Event {
...
+export interface SplunkOtelWebEventTarget {
+  _experimental_addEventListener: InternalEventTarget['addEventListener'];
+  _experimental_removeEventListener: InternalEventTarget['removeEventListener']
+}
 
-  dispatchEvent(event: SessionIdChangedEvent | GlobalAttributesChangedEvent): void {
-    this.target.dispatchEvent(event);
-  }
+export function buildInternalEventTarget(): InternalEventTarget {
+  return document.createElement('div');
```
t2t2, line 60: Maybe a comment for in 3 months when everyone's forgotten about this

```suggestion
  // EventTarget can't be initiated in older versions of safari and IE so use something that already implements it
  return document.createElement('div');
```

t2t2 (approved):

PR #161, opened 2021-05-13
Feat/inline error catcher

jtmalinowski:


docs/Installation.md
```diff
@@ -114,7 +114,11 @@ Trace linkage with APM is currently supported for following Splunk APM agents:
 
 If your configuration is unsupported, you can still stitch the APM trace with the RUM information by [enabling the trace linking manually](https://github.com/signalfx/splunk-otel-js-browser/blob/main/docs/ServerTraceContext.md).
 
+## Hybrid inline / body bottom mode
+
+If you decide not to load the instrumentation synchronously in the HEAD due to its size, but you would like to track of errors happening during the page load, you can add a separate small inline error tracker.
```
johnbley, line 119: A person reading this might ask, "How?  Where?", and in response we could provide more details here or a link to a specialized page outlining the complexities of this approach.

t2t2 (approved):

src/inlineErrorReporter.ts
```diff
@@ -0,0 +1,67 @@
...
+
+declare global {
+  interface Window {
+    __SplunkRumInline?: SplunkRumInlineType;
+  }
+}
+
+const store: ErrorReport[] = [];
+function report(source: ErrorReport['source'], arg: ErrorReport['arg']) {
+  store.push({ source, arg });
```
mhennoch, line 34: I would add some kind of attribute to errors reported by inline reporter so you could tell that it is from inline reporter

jtmalinowski: Done, but something's broken now, because tests added for it are failing

t2t2: passing Event to `report` -> `reportErrorEvent` -> `report` lost the [extraAttrs so added it](https://github.com/signalfx/splunk-otel-js-web/pull/161/commits/11cfafca12eea628d086a008a09b3c5502e7a797#diff-940884dbc6718136929c55cb91e01215c2f1f2b19b3c7d020aacf50780a9d82e), also while testing locally there was quite a lot of racing happening where full agent started before error was registered by inline handler so updated those tests to load full agent after error happened for sure

t2t2: me: for sure
CI: And then, get this, he thought it would work for sure!

PR #166, opened 2021-05-17
Create todolist example

t2t2:
Add example todo app with react frontend and a simple express backend. Includes examples in configuration for ignoring urls (/ping that is on setInterval) and redacting url (id of item)

APMI-1623


examples/todolist/.env.example
```diff
@@ -0,0 +1,3 @@
+REACT_APP_BACKEND=http://localhost:3001
+REACT_APP_RUM_BEACON_URL=https://rum-ingest.<REALM>.signalfx.com/v1/rum
```
jtmalinowski, line 2: I think that we should default to US0 here for ease of use, maybe add a comment that it might need to be changed.

examples/todolist/src/instrumentation.js
```diff
@@ -0,0 +1,35 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import SplunkOtelWeb from '@splunk/otel-web';
+
+SplunkOtelWeb.init({
```
jtmalinowski, line 19: We recommend CDN first, so it'd be confusing if our main example app is not sticking to recommendations.

examples/todolist/src/instrumentation.js
```diff
@@ -0,0 +1,35 @@
...
+  rumAuth: process.env.REACT_APP_RUM_TOKEN,
+  app: 'RUM Todo Example Frontend',
+  // Ignore 30s ping
+  ignoreUrls: [/\/ping/],
+  // Redact ID from url
+  exporter: {
+    onAttributesSerializing: (attributes) => ({
+      ...attributes,
+      'http.url': typeof attributes['http.url'] === 'string'
+        ? attributes['http.url'].replace(/(\/items\/)\d+(\/?)/, '$1<id>$2')
```
jtmalinowski, line 31: This is a bad example, because it confuses the users about the purpose of this mechanism (PII masking) and might make them think that they have to normalise URLs manually. Can we instead do something with a fake oauth token?

examples/todolist/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+This example todo application consists of react frontend app and express backend app.
+
+## Try it yourself!
+
+After downloading this folder:
+
+1. Run `npm install`
+2. Copy `.env.example` to `.env`
+3. Fill out `REACT_APP_RUM_*` with your ingest URL and RUM token
```
jtmalinowski, line 11: Link to trial account registration?

mhennoch (approved):

PR #200, opened 2021-06-30
feat: support for Synthetics-RunId

jtmalinowski:


src/synthetics.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  return propagation.createBaggage({
+    [SYNTHETIC_RUN_ID_FIELD]: { value },
+  });
+}
+
+export function extractSyntheticsAttributesFromContext(context: Context): SpanAttributes {
+  const syntheticsRunId = propagation.getBaggage(context)?.getEntry(SYNTHETIC_RUN_ID_FIELD).value;
+
+  if (!isValidSyntheticsRunId(syntheticsRunId)) {
```
mhennoch, line 68: Could get rid of negation:
```
  if (isValidSyntheticsRunId(syntheticsRunId)) {
     return {
        [SYNTHETIC_RUN_ID_FIELD]: syntheticsRunId,
      }
  }

  return {};

File: src/synthetics.ts [hunk 2]
```diff
@@ -0,0 +1,75 @@
...
+  const baggage = propagation.getBaggage(context) || propagation.createBaggage();
+  const baggageWithSynthetics = baggage.setEntry(SYNTHETIC_RUN_ID_FIELD, { value: syntheticsRunId });
+
+  return propagation.setBaggage(context, baggageWithSynthetics);
+}
+
+export function extractSyntheticsBaggageFromWindow(): Baggage | null {
+  const value = window.syntheticsRunId;
+
+  if (!isValidSyntheticsRunId(value)) {
```
mhennoch, line 56: Same as below, I would try to minimize negations if possible

test/propagation.test.ts
```diff
@@ -0,0 +1,94 @@
...
+      expect(exporter.getFinishedSpans().length).to.eql(1);
+      expect(
+        exporter.getFinishedSpans()[0].attributes['Synthetics-RunId'],
+      ).to.eql(syntheticsTraceId);
+
+      done();
+    });
+  });
+
+  it('must not attach synthetic run id as an attribute', done => {
```
mhennoch, line 77: Maybe clarify that: must not attach **invalid** synthetic run ...

mhennoch:
Discussion from slack:
Why do we even need this HttpBaggagePropagator? 
https://signalfuse.atlassian.net/browse/POR-7571 if i understand this correctly, synthethics already adds that header to all requests . RUM only needs to read the window.syntheticsRunId and add it to all spans as an attribute.

jtmalinowski:
Closing in favor of #217

PR #211, opened 2021-07-06
Delete INSTALLING.md

mhennoch:
Duplicate


t2t2 (approved):
Unfortunately no way to redirect but oh well, such is life with md-s

PR #212, opened 2021-07-07
Move common attributes to resource attributes

t2t2:
Added [relevant resource telemetry sdk attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#telemetry-sdk), some duplicate currently existing, keeping both around for backwards compat.

Zipkin exporter [copies them from resource to output span tags](https://github.com/open-telemetry/opentelemetry-js/blob/3a327ddea7ab46e686beb765cba5bf4f73a90b94/packages/opentelemetry-exporter-zipkin/src/transform.ts#L86) which means to backend there should be no differences (other than a few added tags)


jtmalinowski:
I'd prefer not modifying Zipkin like that, because if we stick to standards, it lets us work across different languages / stacks. Our discussions are also based on based on adherence to standards, e.g. https://github.com/signalfx/gdi-specification/pull/112

jtmalinowski (changes requested):
I'd prefer not modifying Zipkin like that, because if we stick to standards, it lets us work across different languages / stacks. Our discussions are also based on based on adherence to standards, e.g. signalfx/gdi-specification#112

t2t2:
> I'd prefer not modifying Zipkin like that

Uhhh what modifications tho

* This is completely using otel-s default behavior when it comes to zipkin exporter (so no modifications, if anything it removes existing differences (forcing empty resource))
* [Using resources is following the otel standards for sending such data](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service) and should other exporters (specifically otlp/json) be important it will align to how that data structure exports data (instead of it copied to every data otlp/json has it only once at the root level)

test/init.test.ts
```diff
@@ -170,13 +170,17 @@ describe('creating spans is possible', () => {
...
     const exposedSpan = span as tracing.Span;
-    assert.ok(exposedSpan.attributes['splunk.rumSessionId'], 'Checking splunk.rumSessionId');
-    assert.ok(exposedSpan.attributes['splunk.rumVersion'], 'Checking splunk.rumVersion');
     assert.ok(exposedSpan.attributes['location.href'], 'Checking location.href');
-    assert.ok(exposedSpan.attributes['splunk.scriptInstance'], 'Checking splunk.scriptInstance');
-    assert.strictEqual(exposedSpan.attributes['app'], 'my-app');
     assert.strictEqual(exposedSpan.attributes['environment'], 'my-env');
     assert.strictEqual(exposedSpan.attributes.customerType, 'GOLD');
+    // Attributes set on resource that zipkin exporter merges to span tags
+    assert.ok(exposedSpan.resource.attributes['telemetry.sdk.name'], 'Checking splunk.rumSessionId');
```
johnbley, line 177: Copied+pasted assert messages here.

jtmalinowski (approved):
@t2t2 just document our modifications to zipkin exporter please?

mhennoch (approved):

PR #213, opened 2021-07-09
fix: Remove maximum queue size from BatchSpanProcessor

t2t2:
Currently (at default values) if there's more than 40 spans within the 5s batching interval then all of the new ones are dropped, which is pretty low and internal usage hit this limit with ease.

Removing this value, which causes it [to use the default value of 2048](https://github.com/open-telemetry/opentelemetry-js/blob/611a30f19145e7294efdc587efc863410c900604/packages/opentelemetry-core/src/utils/environment.ts#L102). For really spammy span generators the exporter's 100 per 30s per component limit still applies.


mhennoch (approved):
Could also make it configurable but /shrug

jtmalinowski (approved):

PR #219, opened 2021-07-20
Capture visibility events

mhennoch:
Captured as 0 duration spans as these can't be complete spans(due to us not being able to close some of them in some case) and we don't have events. Disabled for now cause we need to make sure that sending them is fine. At some point we probably need to add this info to each span so data can be filtered in UI.

Code in index.ts is moved cause otherwise some things fire at wrong times.

### DIFF
--- a/integration-tests/utils/globals.js
+++ b/integration-tests/utils/globals.js
@@ -73,11 +73,12 @@ module.exports = {
     browser.globals.getLastServerTiming = backend.getLastServerTiming;
     browser.globals.getUrl = (...args) => backend.getUrl(...args).toString();
 
-    browser.globals.emulateTabSwitchingAway = async () => {
-      await browser.execute(() => {
-        Object.defineProperty(document, 'visibilityState', { value: 'hidden', configurable: true });
-        document.dispatchEvent(new Event('visibilitychange'));
-      });
+    browser.globals.emulateTabSwitching = async (visible) => {
+      await browser.execute((hidden) => {
+        Object.defineProperty(document, 'hidden', { value: hidden, configurable: true });
+        Object.defineProperty(document,'visibilityState', { value: hidden ? 'hidden': 'visible', configurable: true });
+        window.dispatchEvent(new Event('visibilitychange'));
+      }, [visible]);
     };
 
     browser.globals.waitForTestToFinish = async () => {
--- a/src/SplunkPageVisibilityInstrumentation.ts
+++ b/src/SplunkPageVisibilityInstrumentation.ts
@@ -19,12 +19,15 @@ import { InstrumentationBase, InstrumentationConfig } from '@opentelemetry/instr
 import { VERSION } from './version';
 
 const MODULE_NAME = 'splunk-visibility';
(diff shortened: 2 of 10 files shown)


src/SplunkPageVisibilityInstrumentation.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export class SplunkPageVisibilityInstrumentation extends InstrumentationBase {
+  unloading: boolean;
+
+  constructor(config: InstrumentationConfig = {}) {
+    super(MODULE_NAME, VERSION, Object.assign({}, config));
+    this.unloading = false;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
```
jtmal-signalfx, line 33: Can this be configured in eslint? Especially, if all instances of `InstrumentationBase` require `init()` to be present.

src/SplunkPageVisibilityInstrumentation.ts
```diff
@@ -0,0 +1,65 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { hrTime } from '@opentelemetry/core';
+import { InstrumentationBase, InstrumentationConfig } from '@opentelemetry/instrumentation';
+import { VERSION } from './version';
+
+const MODULE_NAME = 'splunk-visibility';
+let visibilityListener: any;
```
jtmal-signalfx, line 22: Can these listeners be fields of the instrumentation?

integration-tests/utils/globals.js
```diff
@@ -86,11 +86,12 @@ module.exports = {
...
     Object.assign(browser.globals, await buildBackendContext(browser));
 
-    browser.globals.emulateTabSwitchingAway = async () => {
-      await browser.execute(() => {
-        Object.defineProperty(document, 'visibilityState', { value: 'hidden', configurable: true });
-        document.dispatchEvent(new Event('visibilitychange'));
-      });
+    browser.globals.emulateTabSwitching = async (visible) => {
+      await browser.execute((hidden) => {
+        Object.defineProperty(document, 'hidden', { value: hidden, configurable: true });
```
jtmal-signalfx, line 91: Looking at this, I'd suggest we make it a unit test. This isn't how an integration test should work.

jtmal-signalfx (approved):
No major issues, I'd prefer that the tests for visibility are unit tests because of how they work.

src/SplunkPageVisibilityInstrumentation.ts
mhennoch: Switched off globally

src/SplunkPageVisibilityInstrumentation.ts
mhennoch: ✅

integration-tests/utils/globals.js
mhennoch: Will look into making it a real integration test with a real tab switch at some point

t2t2 (approved):

PR #228, opened 2021-08-04
Offline online

mhennoch:
Capture offline online events as 0 duration spans.


src/SplunkConnectivityInstrumentation.ts
```diff
@@ -0,0 +1,62 @@
...
+      this.offlineStart = hrTime();
+    });
+
+    this.onlineListener = window.addEventListener('online', () => {
+      if (this.offlineStart) {
+        // this could be a span but let's keep it as an "event" for now.
+        this._createSpan(false, this.offlineStart);
+        this._createSpan(true, hrTime());
+      }
+    });
```
t2t2, line 48: addEventListener doesn't return anything so these listeners stay undefined

t2t2 (approved):

PR #241, opened 2021-08-24
Skip export config option

### LINKED ISSUE #240: Config option to disable all exporting (console and network)
My use case is that I'd like to @splunk/otel-web as a local tracer provider / source of auto-instrumentation, but implement my own exporter to send its spans to a different ingest location.

sfishel-splunk:
# Description

Adds config option to disable all exporting so that @splunk/otel-web can be used as a local tracer provider / auto-instrumentation source.

Closes #240

## Type of change

# How has this been tested?

- [ ] Manual testing
- [ ] Added unit tests
- [ ] Added integration tests

# Checklist:

- [ ] Changelogs have been updated
- [ ] Unit tests have been added/updated
- [ ] Documentation has been updated
- [ ] Team's resident technical writer tagged on the PR (if there are changes to the documentation)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ If the version of Open Telemetry is unspecified for a version, then it is the sa
 
 ## Unreleased
 - Fix Internet Explorer compatibility
+- Add `skipExport` configuration option [#240](https://github.com/signalfx/splunk-otel-js-web/issues/240)
 
 ## 0.8.0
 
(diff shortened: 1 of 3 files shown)


jtmal-signalfx (changes requested):
We use `_internalInit` (see <https://github.com/signalfx/splunk-otel-js-web/blob/2340562829e672ef6548ac27673a31a4a2481f1f/src/index.ts#L204>) for the purpose of offering various experimental or internal options. I can't think of a case when our customer would need this option.

Current implementation of `_internalInit` also allows you to replace the internal `Exporter` with your own, which will be a bit better, because we do keep certain functionalities in `SpanProcessor`, due to OTel's architecture. Maybe this option would work for you?

PR #245, opened 2021-08-31
Update CODEOWNERS

theletterf:
Updated so that tech writers can't block PRs.


owais (approved):
I expected the first line to include all of the docs as well

theletterf:
@owais You mean `*`? Not sure if I get your comment.

owais:
Yes, wouldn't `*` match all files?

theletterf:
Indeed. The idea here is that tech writers are included as reviewers for docs, but without exclusive approval power (we don't want tech writers to prevent PRs from going forward if a change in docs is not approved by one of them).

owais:
OK. Either way, it stands approved :)

t2t2 (approved):

PR #304, opened 2022-01-17
Socket.io client instrumentation

t2t2:
# Description

Adds instrumentation to socket.io client that unlike websocket instrumentation follows the [messaging systems](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md) otel spec, as socket.io usage has the structure needed to make it work. It is also made with [aspecto's socket.io server side instrumentation](https://github.com/aspecto-io/opentelemetry-ext-js/tree/master/packages/instrumentation-socket.io) in mind (spans should be similar, tho with a few omissions that don't apply client side).

Since this is the first instrumentation that instruments a separately loaded library (won't always be accessible when RUM is being inited, may not be accessible at all in bundled apps), extra documentation effort was needed to cover multiple possible installation methods.

## Type of change

- Documentation update

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -34,6 +34,7 @@ In order to enable or disable specific instrumentations in the Browser Agent, yo
 |`instrumentations.fetch`|`true`|Enables capturing [fetch requests](Instrumentations.md#instrumentation-xhrfetch) requests.|
 |`instrumentations.interactions`|`true`|Enables capturing [user interactions](Instrumentations.md#instrumentation-user-interactions) (such as clicks or keyboard events).|
 |`instrumentations.longtask`|`true`|Enables capturing [long tasks](Instrumentations.md#instrumentation-long-tasks).|
+|`instrumentations.socketio`|`false`|Enables capturing [socket.io messages](Instrumentations.md#instrumentation-socketio-client).|
 |`instrumentations.postload`|`true`|Enables capturing [resources loaded after load event](Instrumentations.md#instrumentation-post-document-load).|
 |`instrumentations.visibility`|`false`|Enables capturing [visibility events](Instrumentations.md#instrumentation-visibility).|
 |`instrumentations.websockets`|`false`|Enables capturing [websockets](Instrumentations.md#instrumentation-websockets).|
@@ -45,6 +46,7 @@ Some instrumentations, such as interactions module, have additional configuratio
 |Option|Type|Default value|Description|
 |---|---|---|---|
 |`instrumentations.interactions.events`|`{[DOM Event Name]: boolean}`|<pre>{<br>click: true,<br>dblclick: true,<br>mousedown: true,<br>mouseup: true,<br><br>submit: true,<br>reset: true,<br>change: true,<br><br>dragend: true,<br>drop: true,<br><br>ended: true,<br>pause: true,<br>play: true,<br>}<br></pre>|[DOM events](https://developer.mozilla.org/en-US/docs/Web/Events#event_listing) that are captured as user interactions, as captured from `SplunkRum.DEFAULT_AUTO_INSTRUMENTED_EVENTS`|
+|`instrumentations.socketio.target`|`string | socket.io client`|`'io'`|Either the global variable name to which the socket.io client will be loaded, or the io object directly (see the documentation for [socket.io instrumentation](Instrumentations.md#instrumentation-socketio-client))|
 
 ## Changing the configuration: examples
 In situation, where you need to change the default configuration, you need to change the object passed to `SplunkRum.init()` call:
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -46,7 +46,7 @@ Some instrumentations, such as interactions module, have additional configuratio
 |Option|Type|Default value|Description|
 |---|---|---|---|
 |`instrumentations.interactions.events`|`{[DOM Event Name]: boolean}`|<pre>{<br>click: true,<br>dblclick: true,<br>mousedown: true,<br>mouseup: true,<br><br>submit: true,<br>reset: true,<br>change: true,<br><br>dragend: true,<br>drop: true,<br><br>ended: true,<br>pause: true,<br>play: true,<br>}<br></pre>|[DOM events](https://developer.mozilla.org/en-US/docs/Web/Events#event_listing) that are captured as user interactions, as captured from `SplunkRum.DEFAULT_AUTO_INSTRUMENTED_EVENTS`|
-|`instrumentations.socketio.target`|`string | socket.io client`|`'io'`|Either the global variable name to which the socket.io client will be loaded, or the io object directly (see the documentation for [socket.io instrumentation](Instrumentations.md#instrumentation-socketio-client))|
+|`instrumentations.socketio.target`|`string | socket.io client`|`'io'`| Either the global variable name to which the socket.io client will be loaded, or the `io` object. See the documentation for [socket.io instrumentation](Instrumentations.md#instrumentation-socketio-client))|
 
 ## Changing the configuration: examples
 In situation, where you need to change the default configuration, you need to change the object passed to `SplunkRum.init()` call:
--- a/docs/Instrumentations.md
+++ b/docs/Instrumentations.md
@@ -213,7 +213,6 @@ Captures information from `send` and `onmessage` events.
 
 This instrumentation generates spans from messages sent using the [socket.io](https://socket.io/) client library. Generated spans conform to the [OpenTelemetry specifications on messaging systems](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md). This instrumentation is disabled by default.
 
-> Enabling websocket instrumentation when using socket.io isn't generally recommended.
 
 ### Setup
 
@@ -268,6 +267,9 @@ When using the CDN distribution of Splunk RUM, enable the socket.io instrumentat
 </script>
 <script src="/app.min.js"></script>
 ```
+
+
+The content of the `app.min.js` file in this example is the following:
 
 ```js
 import { io } from 'socket.io-client';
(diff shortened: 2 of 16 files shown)


docs/Configuration.md
```diff
@@ -45,6 +46,7 @@ Some instrumentations, such as interactions module, have additional configuratio
 |Option|Type|Default value|Description|
 |---|---|---|---|
 |`instrumentations.interactions.events`|`{[DOM Event Name]: boolean}`|<pre>{<br>click: true,<br>dblclick: true,<br>mousedown: true,<br>mouseup: true,<br><br>submit: true,<br>reset: true,<br>change: true,<br><br>dragend: true,<br>drop: true,<br><br>ended: true,<br>pause: true,<br>play: true,<br>}<br></pre>|[DOM events](https://developer.mozilla.org/en-US/docs/Web/Events#event_listing) that are captured as user interactions, as captured from `SplunkRum.DEFAULT_AUTO_INSTRUMENTED_EVENTS`|
+|`instrumentations.socketio.target`|`string | socket.io client`|`'io'`|Either the global variable name to which the socket.io client will be loaded, or the io object directly (see the documentation for [socket.io instrumentation](Instrumentations.md#instrumentation-socketio-client))|
```
theletterf, line 49:
```suggestion
|`instrumentations.socketio.target`|`string | socket.io client`|`'io'`| Either the global variable name to which the socket.io client will be loaded, or the `io` object. See the documentation for [socket.io instrumentation](Instrumentations.md#instrumentation-socketio-client))|
```

docs/Instrumentations.md
```diff
@@ -208,3 +208,99 @@ Captures information from `send` and `onmessage` events.
 |---|---|---|
 |`http.url`|`string`|Websocket URL.|
 |`response_content_length`|`number`|Payload size in bytes.|
+
+## Instrumentation: Socket.io client
+
+This instrumentation generates spans from messages sent using [socket.io](https://socket.io/) client library. The generated spans conform to [OpenTelemetry specifications on messaging systems](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md). This instrumentation is disabled by default as it requires a bit more advanced setup.
```
theletterf, line 214:
```suggestion
This instrumentation generates spans from messages sent using the [socket.io](https://socket.io/) client library. Generated spans conform to the [OpenTelemetry specifications on messaging systems](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md). This instrumentation is disabled by default.
```

docs/Instrumentations.md
```diff
@@ -208,3 +208,99 @@ Captures information from `send` and `onmessage` events.
...
+
+> Enabling websocket instrumentation when using socket.io isn't generally recommended.
```
theletterf, line 216:
```suggestion
> Enabling websocket instrumentation when using socket.io isn't generally recommended.
```

Can we be more specific about why it shouldn't be done?

docs/Instrumentations.md
```diff
@@ -208,3 +208,99 @@ Captures information from `send` and `onmessage` events.
...
+
+### Setup
+
+#### Socket.io client library served by Socket.io server
+
+When using [the standalone build server by socket.io server](https://socket.io/docs/v4/client-installation/#standalone-build), instrumentation can be just enabled by passing `true` to configuration.
```
theletterf, line 222:
```suggestion
When using [the standalone build by socket.io server](https://socket.io/docs/v4/client-installation/#standalone-build), you can enable instrumentation by passing `true` to the configuration setting, like in the following snippet:
```

docs/Instrumentations.md
```diff
@@ -208,3 +208,99 @@ Captures information from `send` and `onmessage` events.
...
+      socketio: true
+    }
+  });
+</script>
+<script src="/socket.io/socket.io.js"></script>
+```
+
+#### Bundled socket.io client with NPM Splunk RUM installation
+
+When using both `@splunk/otel-web` and `socket.io-client` by bundling together, socket.io client should be passed to the instrumentation via `target` config option:
```
theletterf, line 239:
```suggestion
When using both the `@splunk/otel-web` and the `socket.io-client` packages bundled together, pass the socket.io client to the instrumentation using the `target` config option:
```

docs/Instrumentations.md
```diff
@@ -208,3 +208,99 @@ Captures information from `send` and `onmessage` events.
...
+    socketio: {
+      target: io,
+    },
+  },
+});
+```
+
+#### Bundled socket.io client with CDN Splunk RUM installation
+
+When using CDN distribution of Splunk RUM, socket.io instrumentation should be enabled expecting a global object and inside the bundle `io` function should be exposed as `window.io`
```
theletterf, line 257:
```suggestion
When using the CDN distribution of Splunk RUM, enable the socket.io instrumentation using a global object and exposing the `io` function as `window.io`, as in the following example:
```

Is the rewording correct?

docs/Instrumentations.md
```diff
@@ -208,3 +208,99 @@ Captures information from `send` and `onmessage` events.
...
+  SplunkRum.init({
+    // ...
+    instrumentations: {
+      socketio: true
+    }
+  });
+</script>
+<script src="/app.min.js"></script>
+```
+
```
theletterf, line 271: The JS snippet that follows is the content of what? `app.min.js`?

docs/Instrumentations.md
```diff
@@ -208,3 +208,99 @@ Captures information from `send` and `onmessage` events.
...
+SplunkRum.init({
+  // ...
+  instrumentations: {
+    socketio: {
+      target: 'socketIoClient',
+    },
+  },
+});
+
+// Then expose the io object in your bundle by doing:
```
theletterf, line 293:
```suggestion
// Expose the io object in your bundle
```

docs/Instrumentations.md
```diff
@@ -208,3 +208,99 @@ Captures information from `send` and `onmessage` events.
...
+window.socketIoClient = io;
+```
+
+### Spans
+
+Messages sent between socket.io client and server are sent as `EVENT_NAME send` (client -> server) or `EVENT_NAME receive` (server -> client) spans with the following tags:
```
theletterf, line 299:
```suggestion
Messages sent between socket.io clients and servers are `EVENT_NAME send` (client to server) or `EVENT_NAME receive` (server to client) spans with the following tags:
```

theletterf (changes requested):

docs/Instrumentations.md
t2t2: It's mostly due to the reason this instrumentation was made - websocket is basically just a tcp connection so instrumenting it just shows some data went in, some data came out, so it's kinda useless when a better instrumentation that includes what type of message was sent also exists. But technically there's nothing conflicting if someone does enable both

docs/Instrumentations.md
t2t2: hmm "enable the socket.io instrumentation using a global object" kinda means nothing, maybe

```suggestion
When using the CDN distribution of Splunk RUM, enable the socket.io instrumentation and expose the `io` function as `window.io`, as in the following example:
```

docs/Instrumentations.md
theletterf: When would one prefer websocket over socket.io? It's just because of the messaging data model? If that's the case, why doesn't the websocket instrumentation follow it? Trying to understand if we can replace this note with a more useful recommendation based on instrumentation needs.

docs/Instrumentations.md
t2t2: Yes, or at least the source files that eventually get bundled together into the minified `app.min.js`. Maybe best label for it is "where you use socket.io"

docs/Instrumentations.md
theletterf: +1

docs/Instrumentations.md
t2t2: Hmm so issue that websocket instrumentation has is that it can't really give a good [span name as wanted by the messaging spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#span-name) - since the data can be in any structure as defined by the user (can be a random string `hello`, can be a string with structure that only the dev knows eg `hello/user=1`, can be stringified json, can be raw bytes or [msgpack](https://msgpack.org/), ...) there isn't a good way to extract a name for what data is being sent

Meanwhile socket.io has a low cardinality event name that both client and server use to determine what to do with the message, so we can generate spans with `event_name send/receive` name.

Socket.io uses websocket as once of the transport methods internally so it would cause spans from both instrumentations, where the inner (websocket data sent) span is less useful. Tho socket.io can also do http-polling or comet for sending data when websocket is unavailable, so http spans would be generated then, so I can see removing this warning too since the lower level communication method can appear anyway

theletterf: Then

```suggestion
```

src/utils.ts
```diff
@@ -121,3 +121,45 @@ export function getOriginalFunction<T extends CallableFunction>(func: T): T {
...
+
+  return () => {
+    // Don't touch if value is used or another defineProperty used it
+    if (used || window[identifier] !== value) {
+      return;
+    }
+
+    delete window[identifier];
+    if (value !== undefined) {
+      window[identifier];
```
mhennoch, line 162: `window[identifier]` ?

docs/Instrumentations.md
```diff
@@ -208,3 +208,99 @@ Captures information from `send` and `onmessage` events.
 |---|---|---|
 |`http.url`|`string`|Websocket URL.|
 |`response_content_length`|`number`|Payload size in bytes.|
+
+## Instrumentation: Socket.io client
```
mhennoch, line 212: Maybe mention that we support all versions 1-4.

mhennoch (approved):

theletterf (approved):

PR #364, opened 2022-05-18
make SplunkPostDocLoadResourceInstrumentation aware of upstream conte…

### LINKED ISSUE #357: Feature request: trace parent for post-document-load resource fetch spans
The document-load plugin supports [defining a trace parent for resource fetch spans](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/web/opentelemetry-instrumentation-document-load#optional-send-a-trace-parent-from-your-server), but there is no way to do the same for resource fetches that happen after the document load. 

Proposal: support defining the trace parent with a `data-traceparent` attribute on the `<script>` element that loads the resource, i.e.

```
  <script src="my-script.js" data-traceparent="00-ab42124a3c573678d4d8b21ba52df3bf-d21f7bc17caa5aba-01"></script>
```

If this approach seems reasonable, I'd be happy to submit a PR.

sfishel-splunk:
…xt (#357)

* add a mutation observer for the document head to capture added <script> and  tags,
  maintaining a mapping of the source URLs of those elements to the span context when they were inserted
* use that mapping to look up parent context when creating spans for those resource fetches
* add hooks to SplunkContextManager to ensure that mutation observer records can be processed in the same
  context as the logic that initiates the resource load

# Description

Problem: when a resourceFetch span is created by the SplunkPostDocLoadResourceInstrumentation, context is not propagated from the code that initiated the resource load.

Solution: use a mutation observer to detect when new elements are added to the document head and record the tracing context at that time, to be used later when resourceFetch spans are created.

Fixed #357 

## Type of change

# How has this been tested?

- [x] Manual testing
- [ ] Added unit tests
- [x] Added integration tests

# Checklist:

- [ ] Changelogs have been updated
- [ ] Unit tests have been added/updated
- [ ] Documentation has been updated
- [ ] Team's resident technical writer tagged on the PR (if there are changes to the documentation)


sfishel-splunk:
@t2t2 want to make sure this looks like the right direction, then i'll do another pass of cleanup, adding documentation, etc.

t2t2:
👍 One thing to consider is that MutationObserver doesn't add every child when a tree of nodes is added, so any subchildren (eg once we want to extend to other elements) don't get checked without a deeper iteration

example: https://jsfiddle.net/t2t2/8e06tyLv/

integration-tests/tests/resource-observer/resource-obs.spec.js
```diff
@@ -136,5 +136,52 @@ module.exports = {
     const afterLoadImage = imageSpans.find( span => span.tags['component'] === 'splunk-post-doc-load-resource');
 
     await browser.assert.notEqual(docLoadImage.traceId, afterLoadImage.traceId);
+  },
+  'should propagate tracing context to created spans': async function(browser) {
```
sfishel-splunk, line 140: i see that this suite is currently disabled, but i enabled it locally and everything passed

src/index.ts
```diff
@@ -392,6 +398,9 @@ export const SplunkRum: SplunkOtelWebType = {
     });
 
     this.provider = provider;
+    // TODO: this is so that integration tests can access these APIs, should find a better way
```
sfishel-splunk, line 401: open question here: what's the right way to allow the integration tests to call the `trace` and `context` APIs?

t2t2: Hmm yeah doesn't really have a good way, I'd say grabbing otel object directly from `window[Symbol.for('opentelemetry.js.api.1')]` is good enough for tests

sfishel-splunk: for some reason `window[Symbol.for('opentelemetry.js.api.1')].trace` doesn't seem to have a `setSpan` function. will have to look into that a little more...

t2t2: Yeaah that isn't directly on tracing api but rather uses context (and even then the source for setSpan is in context-utils as a standalone function so it's not really set anywhere)

```js
document.body.addEventListener('click', () => {
  console.log('inspan?', window[Symbol.for('opentelemetry.js.api.1')].context.active().getValue(Symbol.for('OpenTelemetry Context Key SPAN')));
  // setting would be .setValue(Symbol.for(...), span)
});
```

might have to consider creating a separate bundle that exposes otel/api for tests...

mhennoch (approved):

PR #652, opened 2023-09-01
fix(socketio-instrumentation): use apply instead of call method when invoking the 'on' callback

dadu109:
# Description
The `call` method invokes the callback with `args` array as the first argument instead of spreading them, and as a side effect changes the data flow. The `apply` method instead spreads those args into individual arguments for the function.

[Socket.io documentation](https://socket.io/docs/v3/emitting-events/#acknowledgements) shows that data should be passed as it is received and not wrapped in an array

## Type of change

# How has this been tested?

- Manual testing

### DIFF
--- a/packages/web/src/SplunkSocketIoClientInstrumentation.ts
+++ b/packages/web/src/SplunkSocketIoClientInstrumentation.ts
@@ -155,7 +155,7 @@ export class SplunkSocketIoClientInstrumentation extends InstrumentationBase {
             });
 
             try {
-              listener.call(this, args);
+              listener.apply(this, args);
             } catch (error) {
               if (error instanceof Error) {
                 span.recordException(error);


t2t2 (approved):

t2t2:
Even though github will auto-sign the squashed commit, there's a bug which prevents us from merging in case any of the to be merged commits aren't signed. Please sign your commit so we can merge this

* [Github docs: Managing commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification)
* [Stack Overflow: Signing an existing commit with GPG](https://stackoverflow.com/questions/37737096/signing-an-existing-commit-with-gpg)

(Saucelabs tests failing is fine as it requires access to secrets which PRs from forks don't have access to)

dadu109:
Hey @t2t2, got the commit verified 🙌

PR #680, opened 2023-10-11
Update `SessionRecorder` type-definitions to match their use

jzaleski:
# Description

This change-set updates the type-definition and quick-start for `SplunkSessionRecorder` to match the preferred default-configuration.

## Type of change

- Documentation update

# How has this been tested?

- Manual testing


t2t2:
Even though github will auto-sign the squashed commit, there's a bug which prevents us from merging in case any of the to be merged commits aren't signed. Please sign your commit so we can merge this

* [Github docs: Managing commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification)
* [Stack Overflow: Signing an existing commit with GPG](https://stackoverflow.com/questions/37737096/signing-an-existing-commit-with-gpg)

(Saucelabs tests failing is fine as it requires access to secrets which PRs from forks don't have access to, other ci fails can be fixed by updating from base branch)

jzaleski:
@t2t2 I've closed this PR and re-opened a new one here: https://github.com/signalfx/splunk-otel-js-web/pull/684 that is identical to this but also includes an error-case for when the `exportUrl` cannot be determined due to misconfiguration.

PR #745, opened 2024-01-31
Add otlp exporter support behind config option

t2t2:
# Description

Add configuration option `exporter.otlp` that switches from zipkin exporter to otlp/json, while extracting component rate limiter and adding some fixes for using transport based on state of the page (prefer xhr with beacon while unloading)

Updated otel deps as they're back to working condition for web

## Type of change

# How has this been tested?

- Manual testing

Tested by sending data to otel collector. This feature still needs a lot more testing against rum-ingest once that's available 

```js
{
  beaconEndpoint: 'http://localhost:4318/v1/traces',
  exporter: {
    otlp: true,
  },
}
```


packages/web/src/exporters/otlp.ts
```diff
@@ -0,0 +1,70 @@
...
+  constructor(options: SplunkExporterConfig) {
+    super(options);
+    this._onAttributesSerializing = options.onAttributesSerializing || NOOP_ATTRIBUTES_TRANSFORMER;
+  }
+
+  convert(spans: ReadableSpan[]): IExportTraceServiceRequest {
+    // Changes: Add attribute serializing hook to remove data before export
+    spans = spans.map(span => {
+      // @ts-expect-error Yep we're overwriting a readonly property here. Deal with it
+      span.attributes = this._onAttributesSerializing ? this._onAttributesSerializing(span.attributes, span) : span.attributes;
```
mhennoch, line 37: this._onAttributesSerializing alway true no?

packages/web/src/exporters/rate-limit.ts
```diff
@@ -0,0 +1,75 @@
...
+    return this._processor.forceFlush();
+  }
+
+  protected _filter(span: ReadableSpan): boolean {
+    if (span.parentSpanId) {
+      this._parents.set(span.parentSpanId, true);
+    }
+    const component = (span.attributes?.component ?? 'unknown').toString();
+    if (!this._spanCounts.has(component)) {
+      this._spanCounts.set(component, -1);
```
mhennoch, line 47: Why this -1?

packages/web/src/exporters/otlp.ts
t2t2: should be, tho types says maybe

mhennoch (approved):

PR #838, opened 2024-09-17
add source_map_ids to error span attributes

Abinet18:
# Description

retrieve source map ids of files in stacktrace
add source_map_ids to error span attributes

## Type of change

- Documentation update

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests

### DIFF
--- a/packages/web/integration-tests/tests/errors/views/minified_file_errors.ejs
+++ b/packages/web/integration-tests/tests/errors/views/minified_file_errors.ejs
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<title>Errors test</title>
+  <%- renderAgent({}) %>
+</head>
+<body>
+    <h1>Minified files with source map ids</h1>
+    <button id="button1">Button 1</button>
+    <button id="button2">Button 2</button>
+    <script src="/integration-tests/tests/errors/views/script1.min.js" type="module"></script>
+    <script src="/integration-tests/tests/errors/views/script2.min.js"></script>
+</body>
+</html>
--- a/packages/web/integration-tests/tests/errors/views/script1.min.js
+++ b/packages/web/integration-tests/tests/errors/views/script1.min.js
@@ -1,0 +1,18 @@
+/*
+Copyright 2024 Splunk Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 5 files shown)


Abinet18:

packages/web/integration-tests/tests/errors/views/minified_file_errors.ejs
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
t2t2, line 1: Actual integration test (spec.js) missing

packages/web/src/SplunkErrorInstrumentation.ts
```diff
@@ -38,9 +38,33 @@ function stringifyValue(value: unknown) {
...
+    //get sourcemap ids and add to span as error.soruce_map_ids
     span.setAttribute('error.stack', limitLen(err.stack.toString(), STACK_LIMIT));
+    const sourcemapIds = parseErrorStack(err.stack);
+    if (sourcemapIds) {
+      span.setAttribute('error.source_map_ids', sourcemapIds);
+    }
+  }
+}
+
+function parseErrorStack(stack: string): string {
```
t2t2, line 52: `npm run lint:fix`

packages/web/integration-tests/tests/errors/views/minified_file_errors.ejs
```diff
@@ -0,0 +1,15 @@
...
+	<meta charset="UTF-8">
+	<title>Errors test</title>
+  <%- renderAgent({}) %>
+</head>
+<body>
+    <h1>Minified files with source map ids</h1>
+    <button id="button1">Button 1</button>
+    <button id="button2">Button 2</button>
+    <script src="/integration-tests/tests/errors/views/script1.min.js"></script>
+    <script src="/integration-tests/tests/errors/views/script2.min.js"></script>
```
t2t2, line 13: Should also add a test for `<script type="module">` as [it doesn't set document.currentScript](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) and the URL should be gotten from [`import.meta.url` there](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta)

packages/web/integration-tests/tests/errors/views/script1.min.js
```diff
@@ -0,0 +1,5 @@
+document.getElementById("button1").addEventListener("click",function(){throw new Error("Error from script1.js")});
+(function() {
```
t2t2, line 2: Probably should also wrap the content here in try catch block (see issue above about document.currentScript not being set in modules, likely could also end up as an issue if someone does something like `jQuery.getScript`)

t2t2 (approved):

PR #868, opened 2024-10-16
add webpack bundle injection and unit tests to build-plugins

mvirgil:
# Description

This uses the unplugin library, which should help us support other build tools in future.  The "hooks" that
unplugin supports right now does not appear to help with injection into resulting bundles.  So a custom webpack
config was added to help with this.

Adding skipTestLib to unblock work in short-term.  Will need to figure out how to deal with unplugin transitive
dev/peer dependencies which are causing compile issues in its .d.ts file.

## Type of change

# How has this been tested?

- Manual testing
- Added unit tests

Future task will handle integration test

Local manual tests performed using a Cisco repo with a Next.js 14 app


mvirgil:
will replace with new PR

PR #878, opened 2024-11-01
feat: attach `browser.instance.id` and `browser.instance.visibility_state` to spans

Joozty:
# Description

Each span now includes `browser.instance.id`, a unique ID shared across all same-origin frames/contexts within the browser tab. This ID persists across tab reloads until the tab is closed or a different origin is loaded. Additionally, `browser.instance.visibility_state` is added to each span to indicate the tab's visibility state (hidden/visible) at the time the span was created.

## Type of change

# How has this been tested?

- Manual testing
- Added integration tests


Joozty:

Joozty:
recheck

packages/web/src/utils/storage.ts
```diff
@@ -0,0 +1,37 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+export const safelyGetSessionStorage = (key: string): string | null => {
+  let value = null;
+  try {
+    value = window.sessionStorage.getItem(key);
```
jtmalinowski, line 20: okay on the topic of session storage I don't have any strong feelings either way and it seems alright but [following MDN's docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage):

> survives over page reloads and restores.
Page reloads I think we can accept because they'll come up in spans. Restores seem like a rare case. So I think this is okayish?

> Duplicating a tab copies the tab's sessionStorage into the new tab.
Not sure about this one but it also seems rare enough?

I think we can just deploy this and check if it's good enough in practice.

packages/web/src/services/BrowserInstanceService.ts
```diff
@@ -0,0 +1,53 @@
...
+
+
+    // Check if the ID is already stored in the session storage. It might be generated by another frame/context.
+    let browserInstanceId = safelyGetSessionStorage(BROWSER_INSTANCE_ID_KEY);
+    if(!browserInstanceId) {
+      browserInstanceId = generateId(64);
+      safelySetSessionStorage(BROWSER_INSTANCE_ID_KEY, browserInstanceId);
+    }
+    
+    this._id = browserInstanceId;
```
jtmalinowski, line 49: if we didn't manage to persist it then wouldn't it be better to not attach it to spans? Because now let's say someone blocks session storage, and we see their session, it'll have a different browser.instance.id for each `scriptInstance`, tricking you into believing that these are separate windows.

packages/web/src/SplunkSpanAttributesProcessor.ts
```diff
@@ -47,6 +47,7 @@ export class SplunkSpanAttributesProcessor implements SpanProcessor {
     span.setAttribute('location.href', location.href);
     span.setAttributes(this._globalAttributes);
     span.setAttribute('splunk.rumSessionId', getRumSessionId());
+    span.setAttribute('browser.instance.visibility_state', document.visibilityState);
```
jtmalinowski, line 50: nice

packages/web/src/utils/storage.ts
Joozty: > survives over page reloads and restores

I did it intentionally that way. It is representing the browser tab rather than just the current window. I mean it can be adjusted if needed.

>  Duplicating a tab copies the tab's sessionStorage into the new tab.

I guess it is the same as the "restore" case. In the Cisco agent, we cover these use cases, but it requires a bigger refactor as the initialization would need to be asynchronous, and we would need to postpone span sending until the ID is assigned. It can be built on top of this as it is just an additional improvement.

packages/web/src/services/BrowserInstanceService.ts
Joozty: Good point. Please check https://github.com/signalfx/splunk-otel-js-web/pull/878/commits/861bed8b6fa8d6aa53ec419c6baa71fd5f7e7ab6

jtmalinowski (approved):

millcek (approved):

potty (approved):
👍

amertak (approved):

PR #879, opened 2024-11-04
chore: update eslint

Joozty:
# Description

I updated ESLint to the latest version, which introduces a breaking change in the configuration file. The update also adds new recommended rules and integrates Prettier for formatting and consistency. I temporarily disabled some rules (marked as TODO) and will address them later.

## Type of change

Delete options that are not relevant.

- Chore or internal change (changes not visible to the consumers of the package)

# How has this been tested?

Delete options that are not relevant.

Just formatting changes

PR #892, opened 2024-11-12
fix(session-replay): id increment

Joozty:
# Description

Fix bug introduced by the eslint update in #879. Previously, the `++` operator was used, which increments the value after the variable is accessed. We switched to `+=` and moved it before the variable is used, causing issues with the indices.

## Type of change

# How has this been tested?

- Manual testing

### DIFF
--- a/examples/next-ssr-example/README.md
+++ b/examples/next-ssr-example/README.md
@@ -4,7 +4,7 @@ This is a [Next.js](https://nextjs.org/) project bootstrapped using [`create-nex
 
 ## Instrumenting with Splunk RUM using NPM installation method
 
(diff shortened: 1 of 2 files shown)


examples/next-ssr-example/README.md
```diff
@@ -4,7 +4,7 @@ This is a [Next.js](https://nextjs.org/) project bootstrapped using [`create-nex
 
 ## Instrumenting with Splunk RUM using NPM installation method
 
-Next.js uses [static generation](https://nextjs.org/docs/basic-features/pages#pre-rendering) by default to generate the output HTML for most pages. As static generation is done using the node.js runtime, it isn't compatible with RUM. Check that RUM is only imported and activated in the browser environment.
```
Joozty, line 7: The link return 404 and link checker in CI fails. I decided to remove it. This example needs to remove anyway.

amertak (approved):

PR #939, opened 2025-01-27
fix: do not extend session from discarded session replay spans

Joozty:
# Description

This update resolves issues with the `_experimental_allSpansExtendSession` flag, which did not work correctly with the session recorder. Spans created via the session recorder improperly extended sessions in a backgrounded tab.

### Issue:

1. When the session recorder finishes a session in a backgrounded tab, it does not start a new session until the tab returns to the foreground.
2. During this time, the session recorder emits events that are discarded. Before discarding, we call `SplunkRum._internalOnExternalSpanCreated()`, which results in either starting a new session or artificially extending the previous session.

**The changes in this update ensure that the callback is invoked only after confirming that the events will not be discarded.**

### Possible Breaking Changes

The `SplunkRum.getSessionID()` method can return undefined between sessions. It is already typed that way (`string` | `undefined`) but the previous implementation would always return string unless agent is not initialized.

Previously, it would always return a string, even if the session was no longer active, by incorrectly referencing the previous session. **With this change, if there is no activity and no active session, it will return `undefined`.
Spans initiate sessions, so if you call `getSessionID()` between sessions and there is no activity, it will correctly return `undefined`.**

## Type of change

# How has this been tested?

- Manual testing

### DIFF
--- a/packages/session-recorder/src/index.ts
+++ b/packages/session-recorder/src/index.ts
@@ -194,9 +194,7 @@ const SplunkRumRecorder = {
 					return
 				}
 
-				if (SplunkRum._internalOnExternalSpanCreated) {
-					SplunkRum._internalOnExternalSpanCreated()
-				}
+				let isExtended = false
 
 				// Safeguards from our ingest getting DDOSed:
 				// 1. A session can send up to 4 hours of data
@@ -204,6 +202,13 @@ const SplunkRumRecorder = {
 				if (SplunkRum.getSessionId() !== lastKnownSession) {
 					if (document.hidden) {
 						return
+					}
+
+					// We need to call it here before another getSessionID call. This will create a new session
+					// if the previous one was expired
+					if (SplunkRum._internalOnExternalSpanCreated) {
+						SplunkRum._internalOnExternalSpanCreated()
+						isExtended = true
 					}
 
 					lastKnownSession = SplunkRum.getSessionId()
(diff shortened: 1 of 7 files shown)


packages/session-recorder/src/index.ts
```diff
@@ -206,6 +204,13 @@ const SplunkRumRecorder = {
 						return
 					}
 
+					// We need to call it here before another getSessionID call. This will create a new session
+					// if the previous one was expired
+					if (SplunkRum._internalOnExternalSpanCreated) {
```
Joozty, line 209: The hook is called only when we are sure the session recorder chunk is not discarded.

packages/web/src/session/session.ts
```diff
@@ -179,16 +174,15 @@ export function initSessionTracking(
...
 }
 
-export function getRumSessionId(): SessionId | undefined {
+export function getRumSessionId({ useLocalStorage }: { useLocalStorage: boolean }): SessionId | undefined {
 	if (hasNativeSessionId()) {
 		return window['SplunkRumNative'].getNativeSessionId()
 	}
 
-	return rumSessionId
+	return getCurrentSessionState({ useLocalStorage, forceStoreRead: true })?.id
```
Joozty, line 187: When calling `getRumSessionId` we always read the fresh value from storage. Explained in the PR descriptions.

potty (approved):
👍

potty (approved):

amertak (approved):

millcek (approved):

PR #957, opened 2025-02-24
feat: support rrweb and proprietary recorder

potty:
# Description

## Type of change

- Documentation update

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests


.size-limit.cjs
```diff
@@ -36,7 +36,7 @@ module.exports = [
 
 	{
 		name: 'artifacts/splunk-otel-web-session-recorder.js',
-		limit: '22 kB',
+		limit: '101 kB',
```
millcek, line 39: Is this a combination of rrweb and proprietary code? Are there plans to separate them?

millcek (approved):

.size-limit.cjs
potty: Yes, everything is bundled together. We discussed it as the best solution so far, since there aren't any additional scripts, there won't be a problem with CSP.

This is only a temporary solution when users can try the proprietary recorder and in the next version we can remove the rrweb recorder.

amertak (approved):

PR #1039, opened 2025-03-11
sampling session issue

### LINKED ISSUE #1040: TypeError in SessionBasedSampler after upgrading to @splunk/otel-web v0.20.0
After upgrading to `@splunk/otel-web` version `0.20.0`, we are encountering a `TypeError` in the `SessionBasedSampler` component. The error occurs in the `_accumulate` function when attempting to access the `length` property of `sessionId`, which appears to be `undefined`.

Error Message:
```
TypeError
Cannot read properties of undefined (reading 'length')
```

Code Reference:
```ts
    toString() {
        return `SessionBased{ratio=${this._ratio}, sampled=${this._sampled.toString()}, notSampled=${this._notSampled.toString()}}`;
    }
    _accumulate(sessionId) {
        let accumulation = 0;
        for (let i = 0; i < sessionId.length / 8; i++) { // TypeError: Cannot read properties of undefined (reading 'length')
            const pos = i * 8;
            const part = parseInt(sessionId.slice(pos, pos + 8), 16);
            accumulation = (accumulation ^ part) >>> 0;
        }
        return accumulation;
    }
```

Expected Behavior:
The `SessionBasedSampler` should handle cases where `sessionId` is `undefined` gracefully, preventing the TypeError.

Steps to reproduce:
1. Upgrade to `@splunk/otel-web` version `0.20.0`
2. Initialize and use the `SessionBasedSampler`
3. Observe the error when `_accumulate(sessionId)` is called.

jtmalinowski:
# Description

In `0.20.0` when using a SessionBasedSampler there's an error which starts happening after the session expires. To fix that we 

Fixes #1040 

## Type of change

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests


packages/web/src/SessionBasedSampler.ts
```diff
@@ -72,20 +71,22 @@ export class SessionBasedSampler implements Sampler {
...
 		// but replacing deciding based on traceId with sessionId
 		// (not extended from due to private methods)
-		const currentSession = getRumSessionId({ useLocalStorage: this.useLocalStorageForSessionMetadata })
-		if (this._currentSession !== currentSession) {
-			this._currentSessionSampled = this._accumulate(currentSession) < this._upperBound
-			this._currentSession = currentSession
+
+		// TODO: we are guaranteed to have a session here, so maybe error explicitly if that's not the case?
+
+		const currentSessionId = getOrInitShadowSession().id;
```
jtmalinowski, line 83: First we create a shadow session, which we could also call "inactive". Wherever we consider if a session is active (or exists), this `shadow` attribute is checked first.

packages/web/src/SessionBasedSampler.ts
```diff
@@ -72,20 +71,22 @@ export class SessionBasedSampler implements Sampler {
...
+		if (this._currentSessionId !== currentSessionId) {
```
jtmalinowski, line 84: the shadow (or active) session's ID is stable for the same period of time as a normal session would, and session samplers check similarly returns the same result for the same session ID

packages/web/src/SplunkSpanAttributesProcessor.ts
```diff
@@ -43,11 +42,14 @@ export class SplunkSpanAttributesProcessor implements SpanProcessor {
 	}
 
 	onStart(span: Span): void {
+		// now that we have started the span, the session is not a shadow anymore
+		const sessionState = updateSessionStatus({ forceStore: false, shadow: false });
```
Joozty, line 46: we just need to be sure the sampler is called before this hook. I guess it should but I was unable to find it in the otel codebase yet.

packages/web/src/index.ts
```diff
@@ -328,6 +329,7 @@ export const SplunkRum: SplunkOtelWebType = {
 			)
 			return
 		}
+		setStoreType("cookie")
```
Joozty, line 332: you should use persistance flag here

packages/web/src/session/session.ts
```diff
@@ -73,37 +111,46 @@ export function getCurrentSessionState({ useLocalStorage = false, forceStoreRead
...
 	eventTarget?.emit('session-changed', { sessionId: sessionState.id })
 
 	if (recentActivity || forceActivity) {
 		sessionState.expiresAt = Date.now() + SESSION_INACTIVITY_TIMEOUT_MS
-		if (useLocalStorage) {
-			setSessionStateToLocalStorage(sessionState, { forceStoreWrite: shouldForceWrite || forceStore })
-		} else {
-			renewCookieTimeout(sessionState, cookieDomain, { forceStoreWrite: shouldForceWrite || forceStore })
+
+		// TODO: review if this check makes sense
```
Joozty, line 143: I was thinking exactly the same but was unsure if it could be removed. I would go for it, and since there are two of us who think the same, I guess it's safe.

packages/web/src/session/session.ts
```diff
@@ -125,16 +171,8 @@ class SessionSpanProcessor implements SpanProcessor {
...
-			markActivity()
-		}
-
-		context.with(suppressTracing(context.active()), () => {
-			updateSessionStatus({
-				forceStore: false,
-				useLocalStorage: this.options.useLocalStorage,
-			})
-		})
+		// responsibility taken over by SplunkBaseSampler
```
Joozty, line 174: yep please remove it

packages/web/src/storage/cookie-store.ts
```diff
@@ -0,0 +1,92 @@
+import { SESSION_INACTIVITY_TIMEOUT_SECONDS } from "../session/constants";
+import { isIframe } from "../utils";
+import { throttle } from "../utils/throttle";
+
+const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent)
+
+export class CookieStore<T> implements Store<T> {
+	private _cachedValue: T | undefined;
```
Joozty, line 8: We did not enforce it by lint, but we try not to use `_` prefixes. The private keyword is enough. I know this is common in other languages, but if we want to follow JavaScript conventions, we should be prefixing it with #, which looks ugly and does not make sense. Hence, as a team, we decided to just use the private keyword and avoid prefixing.


https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

packages/web/tests/init.test.ts
```diff
@@ -88,6 +88,8 @@ describe('test init', () => {
 			})
 			expect(SplunkRum.inited).toBeTruthy()
 			doesBeaconUrlEndWith(path)
+
+			deinit()
```
millcek, line 92: Could this be moved to `afterEach`?

packages/web/src/session/session.ts
```diff
@@ -53,18 +52,52 @@ export function markActivity(): void {
...
+	const str = getNullableStore()
+
+	if (str) {
+		return str
+	}
+
+	throw new Error('Session store was accessed but not initialised.')
+}
+
+export function getOrInitShadowSession(): SessionState | undefined {
```
amertak, line 79:
```suggestion
export function getOrInitShadowSession(): SessionState {
```

packages/web/tests/init.test.ts
jtmalinowski: done!

millcek (approved):

PR #1065, opened 2025-03-31
fix: do not duplicate spans when load event is triggered manually from code

amertak:
# Description

Fix issue with duplicated spans when load event is triggered manually

## Type of change

# How has this been tested?

- Manual testing
- Added integration tests

### DIFF
--- a/packages/web/src/SplunkPostDocLoadResourceInstrumentation.ts
+++ b/packages/web/src/SplunkPostDocLoadResourceInstrumentation.ts
@@ -67,8 +67,8 @@ export class SplunkPostDocLoadResourceInstrumentation extends InstrumentationBas
 
 	enable(): void {
 		if (window.PerformanceObserver) {
-			window.addEventListener('load', () => {
-				this._startPerformanceObserver()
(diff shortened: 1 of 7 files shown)


packages/web/src/SplunkPostDocLoadResourceInstrumentation.ts
```diff
@@ -67,8 +67,8 @@ export class SplunkPostDocLoadResourceInstrumentation extends InstrumentationBas
 
 	enable(): void {
 		if (window.PerformanceObserver) {
-			window.addEventListener('load', () => {
-				this._startPerformanceObserver()
+			window.addEventListener('load', (e) => {
```
millcek, line 70: Could the load event be fired before the agent is initialized?

millcek (approved):

packages/web/src/SplunkPostDocLoadResourceInstrumentation.ts
amertak: Fixed

PR #1075, opened 2025-04-01
feat: session replay module

potty:
# Description

### Major changes
- SplunkRumRecorder: added `recorder` config option to change recorder type. Allowed values: `rrweb` (default), `splunk`. In the future, `rrweb` will be replaced by `splunk` recorder.
- SplunkRumRecorder config is extended by Splunk recorder options. Some of the rrweb config options are migrated to splunk recorder (privacy options mostly).
- Improved log exporting (keepalive fetch, fetch retries) and added log export queue

### Tooling changes
- Rollup replaced by Webpack

## Type of change

- Documentation update

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests


packages/session-recorder/src/BatchLogProcessor.ts
```diff
@@ -26,47 +26,142 @@ export interface BatchLogProcessorConfig {
...
+
+	private flushAll(): Promise<void> {
+		return new Promise((resolve, reject) => {
+			const promises = []
+			promises.push(this.flushOneBatch())
+
+			Promise.all(promises)
+				.then(() => {
+					resolve()
+				})
```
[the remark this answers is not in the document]
potty: Tady bohuzel rve TS kvuli mismatch `void[]` a `void`, `Promise.all` totiz vraci array a resolve ma arg `void`.

packages/session-recorder/src/OTLPLogExporter.ts
```diff
@@ -126,19 +136,112 @@ export default class OTLPLogExporter {
...
+	}
+}
 
-		fetch(this.config.beaconUrl, {
+const sendByFetch = async (
+	endpoint: string,
+	headers: HeadersInit,
+	data: BodyInit,
+	onSuccess: () => void,
+	keepalive?: boolean,
```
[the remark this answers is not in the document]
potty: Do you mean to use object as arg and consolidate the types with `ApiParams`?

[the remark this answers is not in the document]
millcek: I propose that headers, data/body, and keepalive be consolidated into a single parameter, as these are directly passed. The remaining parameters can either remain unchanged or be included within this parameter as well.

packages/session-recorder/src/BatchLogProcessor.ts
```diff
@@ -26,47 +26,142 @@ export interface BatchLogProcessorConfig {
 }
 
 export class BatchLogProcessor {
```
Joozty, line 28: Ho does it differ from OTel BatchLogProcessor?


https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/sdk-logs/src/platform/browser/export/BatchLogRecordProcessor.ts#L21

potty: It's basically the same, but uses Splunk custom `Log` type instead of `ReadableLogRecord` from OTel.

potty: As we were discussing recently, by using `LogRecord` from OTel would simplify things a lot.

packages/session-recorder/src/session-replay/cdn-module.ts
```diff
@@ -15,22 +15,14 @@
...
-import commonjs from '@rollup/plugin-commonjs'
 
-export const babelPlugin = babel({
-	babelHelpers: 'runtime',
-	extensions: ['.js', '.es6', '.es', 'mjs', '.ts'],
-})
+import {
+	SessionReplay as SessionReplayOriginal,
+	SessionReplayPlain as SessionReplayPlainOriginal,
+	// @ts-expect-error Remote module
```
Joozty, line 22: This approach (.d.ts) does not work to make types available?

declare module '[session-replay.module.legacy.min.js](https://cdn.signalfx.com/o11y-gdi-rum/session-replay/v1.33.0/session-replay.module.legacy.min.js)' {
....
}

packages/session-recorder/src/storage.ts
```diff
@@ -0,0 +1,47 @@
...
+	try {
+		value = window.localStorage.getItem(key)
+	} catch {
+		// localStorage not accessible probably user is in incognito-mode
+		// or set "Block third-party cookies" option in browser settings
+	}
+	return value
+}
+
+export const safelySetLocalStorage = (key: string, value: string): boolean => {
```
Joozty, line 29: We have these functions in agent... we should find a way to share it here to to avoid duplicate code. (Can be done in new PR afterwars)

packages/session-recorder/src/utils.ts
```diff
@@ -29,9 +29,11 @@ const GLOBAL_SPLUNK_RUM_VERSION_KEY = `${GLOBAL_SPLUNK_RUM_KEY}.version`
  * eg. sharing session id manager with session recorder
  */
 export function getGlobal<Type>(): Type | undefined {
+	// @ts-expect-error FIXME: Symbol type is not assignable to string
```
Joozty, line 32: Why it worked before?

Joozty: I see you added strict check 👍

packages/session-recorder/src/storage.ts
potty: I agree

packages/session-recorder/src/session-replay/cdn-module.ts
potty: It works, thanks for the tip!

Joozty (approved):

PR #1083, opened 2025-04-04
fix: report usage in node

millcek:
https://splunk.atlassian.net/browse/SWAT-9205

### DIFF
--- a/packages/web/src/index.ts
+++ b/packages/web/src/index.ts
@@ -15,6 +15,10 @@
  * limitations under the License.
  *
  */
+if (typeof process !== 'undefined' && process.version) {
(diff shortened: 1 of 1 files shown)


packages/web/src/index.ts
```diff
@@ -16,6 +16,10 @@
  *
  */
 
+if (typeof window !== 'object') {
```
jtmalinowski, line 19: aren't `import` statements hoisted above it anyway? AFAIK it needs to be in its own file imported before anything else.

millcek: That's [correct](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#hoisting), thanks.

amertak (approved):

PR #1108, opened 2025-04-29
fix: calling sendBeacon in IE11

millcek:
# Description

The file `packages/web/src/exporters/zipkin.ts` does not contain a sendBeacon check. In IE11, sendBeacon is not defined, so it throws an error.

## Type of change


Joozty (approved):
👍

amertak (approved):

PR #1182, opened 2025-06-16
chore: switch to pnpm

Joozty:
# Description

## Type of change

# How has this been tested?

- Manual testing


packages/web/LICENSE
Joozty, line 1: image

packages/session-recorder/package.json
```diff
@@ -34,7 +34,7 @@
 		"type-fest": "^4.37.0"
 	},
 	"peerDependencies": {
-		"@splunk/otel-web": "^0.23.0-alpha.2"
+		"@splunk/otel-web": "workspace:*"
```
Joozty, line 37: pnpm will replace it with the current version, but we will be missing the ^. I think that's okay, since we strictly require the exact same version to be installed.

packages/build-plugins/tests/integration.test.node.ts
Joozty, line 1: if you want to run tests in node you need to use test.node.ts suffix

potty (approved):
🔥

PR #1197, opened 2025-06-24
feat: new `SplunkRum.reportError(error, context)` API

Joozty:
# Description

A new `reportError` API has been introduced to replace the now-deprecated `SplunkRum.error()` method. This API provides a more flexible way to report errors, allowing you to include optional additional context.

### Key Features

- Accepts a variety of input types: `string`, `Event`, `Error`, `ErrorEvent`, or an array of any values.
- Supports an optional `context` parameter to attach custom context to the error.
- The `context` is a simple object containing key-value pairs.
- These key-value pairs are added to the error span for enhanced monitoring and debugging.

### API Signature

```ts
reportError: (
  error: string | Event | Error | ErrorEvent,
  context?: Record<string, string | number | boolean>,
) => void
```

Use `reportError` to capture errors along with any relevant custom metadata to improve traceability and diagnostics.

## Type of change

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests


jtmalinowski:
Quick thoughts on typing:

```js
  error: /* ... | */ unknown[],
```

why? how would this `unknown[]` come to be? How would be even treat it?

```js
  context?: Record<string, string | number | boolean | string[] | number[] | boolean[]>,
```

We will serialize to `Record<string,string>` anyway, so this seems potentially confusing to the customers, or at least I can see customers who will send `{someIds: [1,2,3]}` and will later ask for a backend filter in `1 in someIds` which we don't currently support.

Joozty:
1. If you report an `unknown[]`, `ErrorInstrumentation` iterates over all arguments, tries to find the first error, and uses it for the stack trace. The error message is composed by stringifying all items in the array.

2. The reason behind it is explained here: https://github.com/signalfx/splunk-otel-js-web/pull/1197#discussion_r2166530859. I don't have a strong opinion about it either, and based on testing, I can confirm that a numeric array gets serialized as `1,2,3`. @amertak  what do you think about Jakub's comment?

Joozty:
Thanks for the feedback, @jtmalinowski . We decided to allow only `Record<string, string | boolean | number>` and discard `unknown[]` from `reportError`.

amertak (approved):

jtmalinowski (approved):

PR #1214, opened 2025-07-08
chore: improvements for session replay

Joozty:
# Description

Things done in this PR:

- Stored logs as plain JSON in localStorage
Logs are stored in plain JSON format for storage efficiency.  Conversion to a Uint8Array happens before send.

- Ensured logs are only removed after successful send
Logs remain in localStorage until we confirm they’ve been successfully sent, to prevent data loss.

- Each API call sends only one segment.
We avoid batching logs here because the underlying SplunkSessionRecorder already batches events internally.

- Stop recording when the browser tab is hidden

## Type of change

# How has this been tested?

- Manual testing


packages/session-recorder/src/index.ts
```diff
@@ -313,6 +313,9 @@ const SplunkRumRecorder = {
 				log.debug('Emitting log', logData)
 
 				processor.onEmit(logData)
+				if (recorderType === 'splunk') {
+					void processor.forceFlush()
```
Joozty, line 317: one segment per API call

PR #1252, opened 2025-07-15
Fix sourceMapIds not found if asset URLs contain a query string

rybon:
# Description

## Type of change

- Documentation update

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests

### DIFF
--- a/packages/web/src/SplunkErrorInstrumentation.ts
+++ b/packages/web/src/SplunkErrorInstrumentation.ts
@@ -30,7 +30,7 @@ import { diag } from '@opentelemetry/api'
 const STACK_LIMIT = 4096
 const MESSAGE_LIMIT = 1024
 
-export const STACK_TRACE_URL_PATTER = /([\w]+:\/\/[^\s/]+\/[^\s?:#]+)/g
+export const STACK_TRACE_URL_PATTER = /([\w]+:\/\/[^\s/]+\/[^\s?:#]+(\?\S[^)]*)?)/g
 
 function useful(s: string) {
 	return s && s.trim() !== '' && !s.startsWith('[object') && s !== 'error'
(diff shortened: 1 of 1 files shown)


rybon:

Abinet18 (changes requested):
This change will not be needed here, it breaks some other tests. This issue needs a fix on the cli tool rather, I have tested the parsing here against the stack trace provided and it is behaving as expected. The source map ids are not expected to contain keys with the query params like this,  'https://www.example.com/_next/static/chunks/webpack-4124fc546f5a1faa.js?dpl=v1.0.0', this mapping is added to the js by cli tool so the fix there should give correct keys

Abinet18:
@rybon , As you mentioned The ErrorInstrumentation parser here strips the query strings while the cli tools parser doesn't. The ErrorInstrumentation is written with such expectation and you can see from the tests here, https://github.com/signalfx/splunk-otel-js-web/blob/412e7c11a4127fc82ad3b9b23dcbcd5ac48f20f2/packages/web/tests/SplunkErrorInstrumentation.test.ts#L127 , I thought it should be the cli tool that needs update but it could also be this one . It may need more discussion, will get back to you after discussing with team.

packages/web/src/SplunkErrorInstrumentation.ts
```diff
@@ -30,7 +30,7 @@ import { diag } from '@opentelemetry/api'
 const STACK_LIMIT = 4096
 const MESSAGE_LIMIT = 1024
 
-export const STACK_TRACE_URL_PATTER = /([\w]+:\/\/[^\s/]+\/[^\s?:#]+)/g
+export const STACK_TRACE_URL_PATTER = /([\w]+:\/\/[^\s/]+\/[^\s?:#]+(\?\S[^)\s*]*)?)/g
```
Abinet18, line 33: To have consistency with cli tool, let us use 
`export const STACK_TRACE_URL_PATTER =/[\w]+:\/\/[^\s]+?(?::\d+)?(?=:[\d]+:[\d]+)/g`

rybon: Alright. So, just change this one line in this PR?

Abinet18:
just change the expected values for test6 at line 132 in  /tests/SplunkErrorInstrumentation.test.ts,
`const expected6 = [
	'http://example.com:8080/path/js/main.js?name=testname',
	'http://example.com:8080/path/js/main2.js#fragmentHere',
	'http://example.com:8080/path/js/main3.js?name=testname#fragmentHere',
]`

rybon:
@Abinet18 done.

Abinet18 (approved):

rybon:
The latest commit seems to have a red build. Could someone restart it?

rybon:
@Abinet18 apparently the build was 11 bytes over the limit. I bumped the threshold to 46 kB. Could you trigger another build?

PR #1253, opened 2025-07-16
chore: implement periodic clearing of expired throttle map items

Joozty:
# Description

To prevent the `throttleMap` from growing indefinitely, we added a periodic task that runs every 5 seconds to remove expired items from the map.

## Type of change

# How has this been tested?

- Manual testing

### DIFF
--- a/.size-limit.cjs
+++ b/.size-limit.cjs
@@ -24,7 +24,7 @@ module.exports = [
 
 	{
 		name: 'artifacts/splunk-otel-web.js',
-		limit: '45 kB',
+		limit: '46 kB',
 		path: './packages/web/dist/artifacts/splunk-otel-web.js',
 	},
 


amertak (approved):

packages/web/src/SplunkErrorInstrumentation.ts
```diff
@@ -217,6 +222,24 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
 		this.endSpanWithThrottle(span, now)
 	}
 
+	private attachClearingInterval(): void {
+		this.clearingIntervalId = setInterval(() => {
+			const callback = () => {
+				this.throttleMap.forEach((relativeTime, throttleKey) => {
+					if (performance.now() - relativeTime > THROTTLE_LIMIT) {
+						this.throttleMap.delete(throttleKey)
```
Abinet18, line 230: Just curious, can a race condition happen here if this is running on interval and sth else is trying to add to the map. Should it be considered ?

Joozty: JavaScript is single-threaded, and this operation is synchronous, so there is no way it would get interrupted by the add operation.

Issue #1276, opened 2025-08-04
Grouped Error IDs for Undescribed Errors in Splunk RUM


aplassen:
Hello,

We’re seeing some recurring and somewhat unclear errors related to elements like <img> and <video>, where there’s no obvious description or indication of what’s causing them. These, along with similar cases, end up grouped under the same error ID in Splunk RUM, which makes it difficult to understand and troubleshoot the underlying issues.
I’ve included a screenshot from a session in Splunk RUM that shows an example.

Context: React 17.0 web application.

Have other customers experienced something similar?
Is this a known limitation or behavior of the RUM JS Agent? And if so, is there a recommended way to gain more insight into these kinds of errors?

<img width="447" height="261" alt="Image" src="https://github.com/user-attachments/assets/42a7d53b-4bbf-440c-bfc6-960b6aed5a9c" />

<img width="560" height="821" alt="Image" src="https://github.com/user-attachments/assets/0bd1a5a2-dfef-4256-adc3-7ed148070068" />

amertak:
Hello,

The presented error is caused by the Error of type Event which can be thrown by an element like image.
The event itself does not contain any useful info except the element that caused it.

When the element is present in the error event, we gather all potential relevant info we can. Right now it is

```
target_element
target_xpath
target_src
```

From your image you can see these attributes. 

<img width="940" height="141" alt="Image" src="https://github.com/user-attachments/assets/20ad1406-6ac5-43ca-8c1e-d452a9c27b20" />

They are pointing to an image that was not probably possible to load (name of the image suggests it).
 
Let me know if there is anything else.

aplassen:
Thanks for the follow-up!

It makes sense given the limitations of onError events in the browser. Even so, I think there’s room for improvement on the RUM side. For example:

Add a message like “Unknown error on <img>”, “Unknown error on <script>”, or more generally “Unknown error on <ELEMENT_TYPE>”. That way, the error is visible directly in RUM without having to dig into sessions, and it might also help prevent those errors from collapsing under a single error ID. I guess it could be handled by the customer via something like https://github.com/signalfx/splunk-otel-js-web/pull/1274 (saw it was closed?), but I’d prefer this to be the standard behaviour.

aplassen:
Any suggestions, comments or insight on the way forward?

Joozty:
@aplassen fixed... we will release a new version this week

PR #1317, opened 2025-08-22
fix: improve error msg for failed resources

Joozty:
# Description

Closes https://github.com/signalfx/splunk-otel-js-web/issues/1276

When the browser fails to load an element — such as an image, video, or script — it triggers an empty event with no message. According to the documentation, this empty error means the resource either failed to load or couldn’t be used. 

New message is introduced saying - `Failed to load <tag_name src="src_attribute_value" />`

https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/error_event

## Type of change

# How has this been tested?

- Added integration tests

### DIFF
--- a/packages/web/src/SplunkErrorInstrumentation.ts
+++ b/packages/web/src/SplunkErrorInstrumentation.ts
@@ -25,6 +25,7 @@ import { isPlainObject, removePropertiesWithAdvancedTypes, SpanContext } from '.
 import { getValidAttributes } from './utils/attributes'
 import { diag } from '@opentelemetry/api'
 import { suppressTracing } from '@opentelemetry/core'
+import { isElement } from './types'
 
 // FIXME take timestamps from events?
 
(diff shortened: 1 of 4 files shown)


packages/web/src/SplunkErrorInstrumentation.ts
```diff
@@ -84,7 +85,7 @@ function parseErrorStack(stack: string): string {
 
 function addStackIfUseful(span: Span, err: Error) {
 	if (err && err.stack && useful(err.stack)) {
-		//get sourcemap ids and add to span as error.soruce_map_ids
+		//g et sourcemap ids and add to span as error.source_map_ids
```
Joozty, line 88:
```suggestion
		// get sourcemap ids and add to span as error.source_map_ids
```

PR #1371, opened 2025-09-15
chore(session-recorder): enforce 2mb limit is log queue

Joozty:
# Description

Add 2MB size limit for queued logs in localStorage

## Type of change

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests


amertak (approved):
👍

PR #1693, opened 2026-03-23
feat: trace separation of xhr and fetch from parent trace like click

Abinet18:
# Description

When a user clicks a button that triggers XHR/Fetch requests, by default all spans (click + XHR/Fetch + backend APM spans) share the same trace ID. This creates one large trace containing everything. This has an impact on BT correlation of rum and apm traces.

With separateTraces: true, each XHR/Fetch request gets its own trace ID. This enables:

Proper Business Transaction (BT) correlation in backend APM
Parent relationship preserved via parent.traceId and parent.spanId attributes for RUM correlation
How It Works
Default behavior (separateTraces: false):

Click Span (traceId: abc123)
  └── XHR Span (traceId: abc123, parentId: click)
        └── Backend Span (traceId: abc123)
With separateTraces: true:

Click Span (traceId: abc123)
 
XHR Span (traceId: xyz789, parent.traceId: abc123, parent.spanId: click)
  └── Backend Span (traceId: xyz789)
The XHR span:

Gets a new trace ID (root span)
Stores the original click span reference in parent.traceId and parent.spanId attributes
Sends its own trace ID in the traceparent header to the backend
Usage
Option 1: Top-level config (applies to both XHR and Fetch)

SplunkRum.init({
  separateTraces: true
})
Option 2: Per-instrumentation config

SplunkRum.init({
  instrumentations: {
    xhr: { separateTraces: true },
    fetch: { separateTraces: true }
  }
})
Option 3: Mixed (top-level with override)

SplunkRum.init({
  separateTraces: true,  // Enable for both
  instrumentations: {
    xhr: { separateTraces: false }  // Override: disable for XHR only
  }
})

## Type of change

- Documentation update

# How has this been tested?

- Added integration tests


packages/integration-tests/src/tests/separate-traces/click-xhr.ejs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    window.testing = true;
+    btnTriggerXhr.addEventListener('click', function () {
+      // Make two XHR requests from within the click handler using test server endpoints
+      makeXHR('/some-data');
+      makeXHR('/no-server-timings');
+      
+      // Wait for spans to be processed
+      setTimeout(function () {
```
Joozty: You probably don’t need these setTimeouts (in all .ejs files) in your tests. You’re already waiting for span elements in the spec files.

Image

packages/web/src/index.ts
```diff
@@ -646,6 +647,13 @@ export const SplunkRum: SplunkOtelWebType = {
 			const instrumentations = INSTRUMENTATIONS.map(({ confKey, disable, Instrument }) => {
 				const pluginConf = getPluginConfig(processedOptions.instrumentations[confKey], pluginDefaults, disable)
 				if (pluginConf) {
+					// Pass top-level separateTraces to XHR and Fetch instrumentations if not overridden
+					if ((confKey === 'xhr' || confKey === 'fetch') && processedOptions.separateTraces !== undefined) {
+						if ((pluginConf as Record<string, unknown>).separateTraces === undefined) {
```
Joozty: you already have in each instrumentation processedOptions so you can get it from there.

Joozty (approved):

PR #1796, opened 2026-05-14
feat: use web-vitals attribution

potty:
# Description

## Summary

Adds experimental Web Vitals attribution support for `webvitals` spans.

The instrumentation now uses `web-vitals/attribution` and can emit additional attribution data for CLS, INP, LCP, and opt-in FCP/TTFB spans. Attribution remains disabled by default, and the default attribution settings avoid exposing sensitive selectors or URL details.

## Changes

- Adds opt-in Web Vitals attribution via `_experimental_attribution`.
- Adds attribution support for CLS, INP, and LCP spans.
- Adds opt-in experimental FCP and TTFB collection.
- Adds privacy controls for element targets and LCP URLs.
- Sanitizes LCP URLs by default and emits safe structural element targets by default.
- Improves Web Vitals reporting behavior around duplicate reports, invalid metric values, and document-load alignment fallback.
- Adds a Web Vitals dev sandbox page for manual attribution testing.
- Expands Playwright coverage for default, attribution-enabled, experimental, and disabled Web Vitals behavior.
- Adds focused Vitest coverage for attribution helpers, sanitization, safe targets, resource timing, span attributes, and instrumentation behavior.
- Adds Chromium unit test coverage reporting in CI.
- Updates the bundle size limit for the web artifact.

## Verification

- Added and updated Vitest unit tests.
- Updated Playwright Web Vitals integration tests.
- CI now runs Chromium unit tests with coverage and uploads coverage artifacts.


packages/web/src/instrumentations/splunk-webvitals-instrumentation.ts
```diff
@@ -16,127 +16,226 @@
...
+		super(MODULE_NAME, VERSION, { ...config, enabled: false })
 
-		this.docLoadSpanPromise = new Promise<ReadableSpan>((resolve) => {
+		this.docLoadSpanPromise = new Promise<ReadableSpan | undefined>((resolve) => {
 			this.docLoadSpanResolve = resolve
 		})
 
+		// Fall back to wall-clock alignment if `documentLoad` never arrives
+		// (e.g. `document` instrumentation is disabled).
+		const fallbackTimeout = setTimeout(() => {
```
Joozty: whats this?

packages/web/src/splunk-context-manager.ts
```diff
@@ -636,14 +636,18 @@ export class SplunkContextManager implements ContextManager {
 				// eslint-disable-next-line @typescript-eslint/ban-ts-comment
 				// @ts-ignore expects __promisify__ for some reason
 				function (this: unknown, ...args: Parameters<typeof setTimeout>) {
+					const originalSetTimeout = original as (
```
Joozty: why do you need this?

potty: Removed the extra cast. It was only a TS workaround and was not needed for runtime behavior.

packages/web/src/instrumentations/splunk-webvitals-instrumentation.ts
potty: Yes, we need this fallback. reportMetric awaits the documentLoad promise when alignment is enabled, so without a timeout a missing documentLoad span would leave the web-vitals metric unreported forever. I clarified the comment and added a unit test that proves we wait only for the bounded grace period before reporting with wall-clock timing.

potty: Removed the timeout after the discussion.

Joozty (approved):

PR #1797, opened 2026-05-15
feat: enable experimental TTFB and FCP webvitals, and additional webvitals attributes

rpd10:
# Description

## Problem Statement
Today, the webvitals instrumentation in `@splunk/otel-web` reports only Core Web Vitals values (CLS, LCP, INP) — single numbers like lcp: 2400. There's no signal in the span explaining details: which DOM element was the LCP target, which interaction caused the worst INP, what shifted to produce the CLS score, or how the number broke down across network/render phases.  Engineers have to reproduce locally and re-instrument to diagnose. 

Additionally, FCP and TTFB — both standard web-vitals metrics — aren't reported at all, leaving gaps in page-load performance visibility.

This change exposes the per-metric attribution data that web-vitals already computes (target selectors, timing breakdowns, load state) and adds opt-in FCP/TTFB span emission, so spans become self-diagnosing without code changes on the consumer side. All behavior is gated behind _experimental_* flags so existing span shape and volume are unaffected for consumers who don't opt in.

## Details
Adds opt-in support for richer web vitals telemetry on @splunk/otel-web's webvitals instrumentation by switching to the web-vitals/attribution subpath import. Three new experimental config flags surface additional data without changing default behavior:
- `_experimental_attribution` — when enabled, every emitted webvitals span gains common attributes (webvitals.name, webvitals.rating, webvitals.navigation_type) plus per-metric attribution attributes describing what caused the value (e.g. lcp.target, lcp.url, lcp.element_render_delay, cls.largest_shift_target, inp.interaction_type, inp.input_delay, etc.)
- `_experimental_fcp` — opt into First Contentful Paint as a webvitals span (with FCP attribution attributes when _experimental_attribution is also enabled). Accepts `boolean | ReportOpts`, mirroring the existing cls/inp/lcp shape.
- `_experimental_ttfb` — opt into Time to First Byte as a webvitals span (with TTFB attribution attributes when _experimental_attribution is also enabled). Same `boolean | ReportOpts` shape.

All three flags default to false, so existing consumers see identical span shape and volume. The _experimental_* prefix follows the project's established lifecycle convention (`_experimental_adjustSessionStartToTimeOrigin`, `_experimental_discardDataAfterInactivity`, etc.); the prefix will be dropped and the default flipped in a future minor release.

The span name remains `webvitals` and the existing per-metric value attribute (tags.lcp, tags.cls, tags.inp) is unchanged, preserving backward compatibility with consumer dashboards and the existing integration tests.

## Type of change

# How has this been tested?

- Added unit tests
- Added integration tests

### DIFF
--- a/packages/integration-tests/src/tests/webvitals/webvitals-additional-metrics.ejs
+++ b/packages/integration-tests/src/tests/webvitals/webvitals-additional-metrics.ejs
@@ -1,0 +1,35 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<title>Webvitals additional metrics</title>
+
+	<%- renderAgent({debug: true, instrumentations: {webvitals: {_experimental_attribution: true, _experimental_fcp: true, _experimental_ttfb: true}}}) %>
+
+	<style>
+		#container {
+			height: 200px;
+			width: 400px;
+		}
+	</style>
+</head>
+<body>
+	<h1>Webvitals additional metrics</h1>
+
+	<a href="javascript:clicky()" id="clicky">Click me</a>
+
+	<div id="container">Don't shift me : (</div>
+	<script>
+		let id = 1;
+		const container = document.getElementById('container');
+		function clicky() {
+			const p = document.createElement('p');
+			p.setAttribute('id', 'p'+id);
+			id++;
+			p.innerText = 'Very long paragraph to force a visual change.  '.repeat(30);
+			container.insertBefore(p, container.firstChild);
+		}
+		window.addEventListener('load', clicky);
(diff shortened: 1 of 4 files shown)


amertak:
Hi @rpd10,

Thank you for your proposed changes.

We are already working on adding attribution to standard metrics in the https://github.com/signalfx/splunk-otel-js-web/pull/1796
 
We will incorporate the proposed experimental FCP and TTFB metrics into that PR as well.

rpd10:

Joozty:
Hi we merged suggested changes in https://github.com/signalfx/splunk-otel-js-web/pull/1796. Thanks for your contribution

PR #1802, opened 2026-05-20
feat: regex support

amertak:
# Description

Add support to specify ignoreUrl as a string in a format `regex/foo/` where `foo` is a Regex. This support is added due to ability to specify config in a JSON.

## Type of change

# How has this been tested?

- Added unit tests
- Added integration tests


Joozty (approved):
👍

PR #1811, opened 2026-05-26
feat: add loaf instrumentation

potty:
# Description

Adds LOAF instrumentation.

image
image

## Type of change

- Documentation update

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests


packages/integration-tests/src/tests/loaf/loaf.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+	if (!isSupported) {
+		test.skip()
+	}
+}
+
+function getLoafSpans(recordPage: RecordPage) {
+	return recordPage.receivedSpans.filter((span) => span.name === LOAF_SPAN_NAME)
+}
+
+function expectNumericAttribute(attributes: Record<string, unknown>, name: string): void {
```
Joozty: Can you maybe extend our matcher. I feel like these functions can be useful everywhere in our integration tests


See
https://github.com/signalfx/splunk-otel-js-web/blob/main/packages/integration-tests/src/utils/expect-span.ts#L125

```
expect(span).toHaveNumericAttribute('loaf.style_and_layout_start')
````

File: packages/web/README.md [hunk 2]
```diff
@@ -177,14 +177,36 @@ Choose a versioning strategy based on your needs:
...
+
+LoAF spans include frame timing attributes such as `loaf.duration`, `loaf.blocking_duration`, `loaf.paint_time`, `loaf.presentation_time`, `loaf.render_start`, `loaf.style_and_layout_start`, and `loaf.first_ui_event_timestamp`. They also include bounded script attribution: `loaf.script_count` reports the original number of scripts, while only the top three scripts by duration are exported as `loaf.script[0..2].*` attributes, including script timing fields such as `duration`, `start_time`, `execution_start`, `pause_duration`, `forced_style_and_layout_duration`, and `source_char_position`. Script source URLs have query strings and fragments stripped for HTTP(S) URLs; non-HTTP values such as `blob:`, `data:`, empty strings, and `<anonymous>` are preserved as reported by the browser.
+
+To bound payload volume, the SDK emits at most 1000 LoAF spans per page session, with up to 50 spans per
```
Joozty: 1000 load per session is low... I would keep only rate limiting,  wdyt?

Also probably add here short explanation why we are dropping load (noise) and that we are scoping them, to reduce just noise from particular scripts while preserving other meaningful loafs. 

PS: even 1000 loaf per session is not correct as you can open multiple tabs or close and reopen tab.

packages/web/src/instrumentations/splunk-long-task-instrumentation.ts
```diff
@@ -90,10 +99,12 @@ export class SplunkLongTaskInstrumentation extends InstrumentationBase {
...
-	private isSupported() {
+	private isSupported(): boolean {
 		// note: PerformanceObserver.supportedEntryTypes has better browser support than LongTask
-		const supportedEntryTypes = window.PerformanceObserver && PerformanceObserver.supportedEntryTypes
-		const effectiveEntryTypes = supportedEntryTypes || []
-		return effectiveEntryTypes.includes(LONGTASK_PERFORMANCE_TYPE)
+		return (window.PerformanceObserver?.supportedEntryTypes ?? []).includes(LONGTASK_PERFORMANCE_TYPE)
+	}
+
+	private shouldSuppressForLoaf(): boolean {
```
Joozty, line 107: 👍

packages/web/src/instrumentations/loaf/types.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+export type PerformanceScriptTimingStable = PerformanceEntry & {
```
Joozty: do you need this?

why you cannot use PerformanceLongAnimationFrameTiming, PerformanceScriptTiming from typescript dom types?

can you add comment is your custom types are needed and explain why, pls?

packages/web/src/instrumentations/splunk-long-animation-frame-instrumentation.test.ts
```diff
@@ -0,0 +1,399 @@
...
+		firstUIEventTimestamp: 0,
+		name: 'long-animation-frame',
+		paintTime: 0,
+		presentationTime: 0,
+		renderStart: 0,
+		scripts,
+		startTime,
+		styleAndLayoutStart: 0,
+		toJSON: () => ({}),
+	} as unknown as PerformanceEntry
```
Joozty: Please avoid unknown cast even in test. It should be used as last resort and ideally explained why it is needed. (types are from web-vitals library)

wouldn't this work?

```ts
function createLoafEntry({
	duration = 161.381,
	scripts = [createScript()],
	startTime = 527.300_000_011_920_9,
}: {
	duration?: number
	scripts?: PerformanceScriptTiming[]
	startTime?: number
} = {}): PerformanceLongAnimationFrameTiming {
	return {
		blockingDuration: 111.4,
		duration,
		entryType: 'long-animation-frame',
		firstUIEventTimestamp: 0,
		name: 'long-animation-frame',
		renderStart: 0,
		scripts,
		startTime,
		styleAndLayoutStart: 0,
		toJSON: () => ({}),
	}
}

function createScript({
	duration = 10,
	executionStart = 0,
	forcedStyleAndLayoutDuration = 0,
	invoker = 'event-listener',
	invokerType = 'event-listener',
	pauseDuration = 0,
	sourceCharPosition = 0,
	sourceFunctionName = 'handler',
	sourceURL = 'https://example.com/app.js',
	startTime = 0,
	windowAttribution = 'self',
}: Partial<{
	duration: number
	executionStart: number
	forcedStyleAndLayoutDuration: number
	invoker: string
	invokerType: ScriptInvokerType
	pauseDuration: number
	sourceCharPosition: number
	sourceFunctionName: string
	sourceURL: string
	startTime: number
	windowAttribution: ScriptWindowAttribution
}> = {}): PerformanceScriptTiming {
	return {
		duration,
		entryType: 'script',
		executionStart,
		forcedStyleAndLayoutDuration,
		invoker,
		invokerType,
		name: 'script',
		pauseDuration,
		sourceCharPosition,
		sourceFunctionName,
		sourceURL,
		startTime,
		toJSON: () => ({}),
		windowAttribution,
	}
}
```

packages/web/src/instrumentations/loaf/support.test.ts
```diff
@@ -0,0 +1,60 @@
...
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { LONG_ANIMATION_FRAME_PERFORMANCE_TYPE } from './constants'
+import { isLoafInstrumentationEnabled, isLongAnimationFrameSupported } from './support'
+
+afterEach(() => {
+	vi.unstubAllGlobals()
+})
+
+describe('LoAF support detection', () => {
```
Joozty, line 29: You should probably also add on top of this test suite a test that runs against all browsers in our unit test matrix and uses the real implementation.

For example:

```
chrome: true
chromium: true
edge: true
firefox: false
webkit: false
```

That would also give us a clear signal when LOAF support lands in additional browsers.

packages/web/src/instrumentations/loaf/source-url.test.ts
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ *
+ */
+
+import { describe, expect, it } from 'vitest'
+
+import { normalizeLoafInvoker, normalizeLoafSourceUrl } from './source-url'
+
+describe('LoAF source URL normalization', () => {
+	it('strips query strings and fragments from http URLs', () => {
```
Joozty: I do not think we need this. We have a global hook called onAttributesSerializing and if customer wants to strip something they should do it there. We should report all what we get.

packages/integration-tests/src/tests/loaf/loaf.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+		expectNumericAttribute(attributes, 'loaf.duration')
+		expectNumericAttribute(attributes, 'loaf.blocking_duration')
+		expectNumericAttribute(attributes, 'loaf.render_start')
+		expectNumericAttribute(attributes, 'loaf.style_and_layout_start')
+		expectNumericAttribute(attributes, 'loaf.first_ui_event_timestamp')
+		expectNumericAttribute(attributes, 'loaf.script_count')
+
+		const scriptCount = Number.parseFloat(String(attributes['loaf.script_count']))
+		expect(scriptCount).toBeGreaterThan(0)
+		for (let index = 0; index < Math.min(scriptCount, 3); index += 1) {
```
Joozty: you should also check if scritpCount is less than or equal to your limit....  and not using `Math.min(scriptCount, 3`

packages/integration-tests/src/tests/loaf/loaf.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+
+		await recordPage.goTo('/long-task/index.ejs')
+		await recordPage.locator('#btnLongtask').click()
+		await recordPage.waitForSpans((spans) => spans.some((span) => span.name === 'longtask'))
+
+		expect(recordPage.receivedSpans.filter((span) => span.name === 'longtask').length).toBeGreaterThanOrEqual(1)
+	})
+})
+
+async function skipIfLoafUnsupported(browserName: string, recordPage: RecordPage): Promise<void> {
```
Joozty: you already have your own isSupported function you can use it here.. you have dedicated unit tests written for it so you can leverage it and avoid duplicate code

packages/web/src/instrumentations/loaf/rate-limit.test.ts
```diff
@@ -0,0 +1,250 @@
...
+		presentationTime: 0,
+		renderStart: 0,
+		scripts,
+		startTime,
+		styleAndLayoutStart: 0,
+		toJSON: () => ({}),
+	}
+}
+
+function createScript({
```
Joozty: you already have 4 functions like this

Image

packages/web/src/instrumentations/loaf/span-attributes.ts
```diff
@@ -0,0 +1,67 @@
...
+	})
+}
+
+function setNumberAttribute(span: Span, name: string, value: number): void {
+	if (isFiniteNumber(value)) {
+		span.setAttribute(name, Math.round(value * 100) / 100)
+	}
+}
+
+function setStringAttribute(span: Span, name: string, value: string | undefined): void {
```
Joozty: can you please extract this to some common global span util as we already have same here https://github.com/signalfx/splunk-otel-js-web/blob/main/packages/web/src/instrumentations/webvitals/span-attributes.ts

packages/web/src/instrumentations/loaf/types.ts
potty: used types from webvitals lib or derived types from them

packages/web/src/instrumentations/loaf/source-url.test.ts
potty: removed sanitization/normalization

Joozty (approved):

PR #1816, opened 2026-06-01
feat: mark page load metrics interrupted on page hide

Joozty:
# Description

End PCT calculation when the document is being unloaded so pending document load and route change spans are still sent before the tab is closed or the page is navigated away from. If PCT calculation is interrupted by pagehide, the span is ended with `browser.navigation.status` set to `interrupted`.

image

## Type of change

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests

### DIFF
--- a/packages/web/src/instrumentations/splunk-user-interaction-instrumentation.ts
+++ b/packages/web/src/instrumentations/splunk-user-interaction-instrumentation.ts
@@ -19,7 +19,12 @@
 import { diag, Span, trace, Tracer, TracerProvider } from '@opentelemetry/api'
 import { isUrlIgnored } from '@opentelemetry/core'
 
-import { SessionManager, SpaMetricsManager } from '../managers'
+import {
+	BROWSER_NAVIGATION_PAGE_COMPLETION_TIME_ATTRIBUTE,
+	BROWSER_NAVIGATION_STATUS_ATTRIBUTE,
+	SessionManager,
+	SpaMetricsManager,
+} from '../managers'
 import { SplunkOtelWebConfig } from '../types'
 import { UserInteractionInstrumentation } from '../upstream/user-interaction/instrumentation'
 import { UserInteractionInstrumentationConfig } from '../upstream/user-interaction/types'
@@ -197,15 +202,19 @@ export class SplunkUserInteractionInstrumentation extends UserInteractionInstrum
 		if (this.spaMetricsManager) {
 			// Wait for all in-flight resources (XHR, fetch, media) to finish loading,
 			// then resolve after a quiet period with no new network activity.
-			const { pct, timeout } = await this.spaMetricsManager.waitForPageLoad({
+			const { pct, status } = await this.spaMetricsManager.waitForPageLoad({
 				startTime: performance.now(),
 			})
 
-			span.setAttribute('browser.navigation.page_completion_time', pct)
-			if (timeout) {
-				span.setAttribute('browser.navigation.status', 'timeout')
+			span.setAttribute(BROWSER_NAVIGATION_PAGE_COMPLETION_TIME_ATTRIBUTE, pct)
+			if (status) {
+				span.setAttribute(BROWSER_NAVIGATION_STATUS_ATTRIBUTE, status)
 			}
 
+			diag.debug('Sending routeChange span with PCT result', {
+				pct,
+				status,
+			})
 			span.end(now + pct)
 			diag.debug('Route change span ended', { pct, span })
 		} else {
--- a/packages/web/src/managers/spa-metrics-manager/quiet-period-awaiter.ts
+++ b/packages/web/src/managers/spa-metrics-manager/quiet-period-awaiter.ts
@@ -18,10 +18,21 @@
 
 import { diag } from '@opentelemetry/api'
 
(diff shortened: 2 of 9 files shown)


packages/web/src/instrumentations/splunk-user-interaction-instrumentation.ts
```diff
@@ -197,15 +197,22 @@ export class SplunkUserInteractionInstrumentation extends UserInteractionInstrum
 		if (this.spaMetricsManager) {
 			// Wait for all in-flight resources (XHR, fetch, media) to finish loading,
 			// then resolve after a quiet period with no new network activity.
-			const { pct, timeout } = await this.spaMetricsManager.waitForPageLoad({
+			const { interrupted, pct, timeout } = await this.spaMetricsManager.waitForPageLoad({
 				startTime: performance.now(),
 			})
 
 			span.setAttribute('browser.nabigation.page_completion_time', pct)
```
Abinet18: typo nabigation, may be constants can be used for these attributes

packages/web/tests/init.test.ts
```diff
@@ -846,6 +881,32 @@ describe('test route change spa metrics timeout', () => {
...
+		try {
+			history.pushState({}, 'title', '/pctInterrupted#WithAHash')
+			window.dispatchEvent(new Event('pagehide'))
+
+			await vi.waitFor(
+				() => {
+					const span = capturer.spans.find((s) => s.name === 'routeChange')
+					expectDefined(span, 'Check if routeChange span is present.')
+					expect(span).toHaveSpanAttribute('browser.navigation.status', 'interrupted')
+					expect(span).toHaveSpanAttribute('browser.nabigation.page_completion_time')
```
Abinet18: typo

packages/web/src/managers/spa-metrics-manager/quiet-period-awaiter.ts
```diff
@@ -77,10 +83,12 @@ export class QuietPeriodAwaiter {
 			const pct = Math.max(this.maxPageLoadWaitTime, 0)
 			diag.debug('QuietPeriodAwaiter: Max page load wait time expired', { pct })
 			this.resolveOnce({
+				interrupted: false,
 				pct,
 				timeout: true,
 			})
 		}, maxPageLoadWaitTime)
+		window.addEventListener('pagehide', this.interruptListener, INTERRUPT_LISTENER_OPTIONS)
```
Abinet18, line 95: Should SpaMetricsManager.stop() dispose the active QuietPeriodAwaiter? This constructor adds a pagehide listener and max-wait timer, but they are only cleaned up from resolveOnce(). If SpaMetricsManager.stop() runs while PCT is still waiting on a slow resource, the awaiter can outlive teardown and later resolve on pagehide/max timeout, potentially ending spans after deinit.

packages/web/src/managers/spa-metrics-manager/quiet-period-awaiter.ts
```diff
@@ -20,8 +20,14 @@ import { diag } from '@opentelemetry/api'
...
 const DEFAULT_MAX_PAGE_LOAD_WAIT_TIME = 180_000
 const DEFAULT_QUIET_TIME = 1000
+const INTERRUPT_LISTENER_OPTIONS: AddEventListenerOptions = { capture: true, once: true }
+const INTERRUPT_LISTENER_REMOVE_OPTIONS: EventListenerOptions = { capture: true }
 
-export type PageLoadMetricsResult = { pct: number; timeout: boolean }
+export type PageLoadMetricsResult = {
+	interrupted: boolean
+	pct: number
+	timeout: boolean
```
Abinet18: It could also be just status instead of two booleans

packages/web/src/instrumentations/splunk-user-interaction-instrumentation.ts
```diff
@@ -197,15 +197,22 @@ export class SplunkUserInteractionInstrumentation extends UserInteractionInstrum
...
-			if (timeout) {
+			if (interrupted) {
```
Abinet18: this is same logic as in doc-load for the new attributes, could use a helper ?

Abinet18 (changes requested):

Abinet18:
Just a thought, are we also planning to report the user_interaction_extended_pct in status ? Can we also report successful status, with sth like 'completed' , It will be easier to filter successful metrics(on the backend/ui)  if added

packages/web/src/managers/spa-metrics-manager/quiet-period-awaiter.ts
Joozty: Yes for the correctness. Not sure if anyone ever calls stop on agent. Anyway I added it thx

packages/web/src/instrumentations/splunk-user-interaction-instrumentation.ts
Joozty: I would prefer it as is. You are right but I think it does not matter that much. If you insist let me know and I will do this change.

potty (approved):

Abinet18 (approved):

PR #1847, opened 2026-06-10
feat: ignore previous-page resources in PCT

Joozty:
# Description

This PR improves SPA Page Completion Time (PCT) accuracy by making each page calculation focus on resources that belong to that page. Previously, a slow request or media load started on one route could continue to delay PCT for the next route. With the new default behavior, pending resources discovered on a previous page are cleared when a new page’s PCT calculation starts.

This helps customers get cleaner page-level performance measurements in SPAs, especially when users navigate while background requests, polling, lazy media, or slow previous-page resources are still in flight. Customers who rely on the previous carry-forward behavior can opt out with `spaMetrics.clearLoadingResourcesOnNewPage: false`.

## Type of change

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests


amertak (approved):

packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts
```diff
@@ -245,7 +250,11 @@ export class SpaMetricsManager {
 				return
 			}
 
-			this.loadingResources.set(event.id, { monitorType: event.monitorType, url: event.url })
+			this.loadingResources.set(event.id, {
+				monitorType: event.monitorType,
+				pageUrl: location.href,
+				url: event.url,
+			})
```
scheler, line 257: From Codex:
Old-page resources are only cleared in waitForPageLoad(), but current-page resources can be discovered before that runs, especially for hash navigations where _emitRouteChangeSpan is triggered by the later hashchange event. If the old page already filled maxResourcesToWatch, the new resource hits this limit and is dropped permanently, so the new page’s PCT can complete too early. I’d clear stale resources with the active config before the maxResourcesToWatch check, or otherwise clear on URL change before accepting/rejecting new discoveries.

Suggested test shape: set maxResourcesToWatch: 1, discover a resource on #old, change to #new without calling waitForPageLoad, discover a #new resource, then call waitForPageLoad; the new resource should still be tracked.

Joozty: I decided to ignore this codex review point. It describes an edge case that would only occur if several fetch requests were triggered immediately after a hash change (synchronous), which is not how single-page applications typically behave.

I have few tests for similar thing (associating the resource to correct url) but decided to leave maxResources as is.

PR #1859, opened 2026-06-23
chore: drive pct network tracking from span events

Joozty:
# Description

This change makes SPA/PCT network tracking use the actual fetch/XHR spans instead of separately patching fetch and XMLHttpRequest.
**Purpose:**
Make PCT network activity align with the spans users see.
Avoid duplicate network-wrapping logic inside SPA metrics.
Track fetch/XHR start and end from span lifecycle events.
Disable/warn when network PCT is configured but fetch/XHR instrumentation is off, because there would be no spans to track.
Improve timing correctness by using span/resource timestamps and dropping stale events from previous page-load windows.
In short: it makes PCT network tracking span-driven and more explainable.

## Type of change

- Documentation update

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests


Abinet18:
stale, not going this direction for tracking span life cycle, revisit if needed

PR #1872, opened 2026-07-06
fix:Track more resource types in post-doc-load-resource-instrumentation

Abinet18:
# Description

This PR expands post-document-load resource span coverage in the browser agent.

Previously, the post-document-load resource instrumentation only tracked `img` and `script` resources by default, and the DOM mutation context mapping only handled elements with a `src` attribute. That meant resources such as CSS, fonts, links, and iframes loaded after document load could appear in browser performance entries but not produce corresponding `resourceFetch` spans.

The change updates the default allowed resource initiator types to include `css`, `font`, `iframe`, `img`, `link`, and `script`, and adds URL extraction for both `src` and `href` based resource elements. This lets post-load CSS and font resources produce `resourceFetch` spans with the existing `splunk-post-doc-load-resource` component.

Fixes: DEMRUM-5893

## Type of change

# How has this been tested?

- Manual testing
- Added integration tests

Tested with:

- `pnpm --dir packages/integration-tests test -- src/tests/resource-observer/resource-observer.spec.ts --project=chromium`

Manual validation was also done through the integration-test fixture page, confirming post-load CSS and font resources now emit spans.
Screenshot 2026-07-06 at 9 22 06 AM


packages/integration-tests/src/tests/resource-observer/resource-observer.spec.ts
```diff
@@ -24,6 +24,19 @@ test.describe('resource observer', () => {
 		await recordPage.goTo('/resource-observer/resources.ejs')
 
 		await recordPage.waitForSpans((spans) => spans.filter((span) => span.name === 'guard-span').length === 1)
+		await recordPage.waitForSpans(
```
shumone: Just curious, for this test, is this waiting for the page to have a span w/ those attributes for css/fonts? Is there a reason why this was needed (like they have an extra delay, or the line 26 await didn't handle it properly?)

packages/integration-tests/src/tests/resource-observer/resources.ejs
```diff
@@ -39,6 +39,29 @@
       return p;
     }
 
+    function loadStylesheetWithFont() {
+      const p = new Promise(function(resolve, reject) {
+        const link = document.createElement('link');
+        link.rel = 'stylesheet';
+        link.href = '/docload/assets/css-font-img.css?postload=true';
```
shumone: Nit: Maybe duplicate the resources to have these in the same package's assets?

packages/web/src/instrumentations/splunk-post-doc-load-resource-instrumentation.ts
```diff
@@ -142,14 +155,13 @@ export class SplunkPostDocLoadResourceInstrumentation extends InstrumentationBas
...
 			.forEach((node) => {
-				const src = node.getAttribute('src')
-				if (!src) {
+				const resourceUrl = getNodeResourceUrl(node)
+				if (!resourceUrl) {
 					return
 				}
 
-				const srcUrl = new URL(src, location.origin)
+				const srcUrl = new URL(resourceUrl, location.origin)
```
mvirgil: Codex review comment:

[P2] [Resolve resource URLs against the document base](https://github.com/signalfx/splunk-otel-js-web/blob/336370100def62cdc3fbaa3e2a3e883846f65516/packages/web/src/instrumentations/splunk-post-doc-load-resource-instrumentation.ts#L159-L165). getAttribute() returns relative values, but new URL(resourceUrl, location.origin) resolves assets/style.css to /assets/style.css, while the performance entry from /resource-observer/page is /resource-observer/assets/style.css. The context lookup misses, producing a root span. Use the element’s resolved .src/.href or document.baseURI

mvirgil (approved):

PR #1873, opened 2026-07-08
feature: Add relevant page span id to attributes of span in the pct window

Abinet18:
# Description

This PR adds browser.navigation.page_span_id to spans that contribute to page completion timing, so analysis can identify which documentLoad or routeChange span a related span contributed to.
The implementation now sets this attribute directly from the relevant instrumentations while spans are still mutable:
document-load documentFetch / resourceFetch
post-document-load resource spans
fetch spans
XHR spans
longtask spans
LoAF spans

## Type of change

# How has this been tested?

- Added integration tests


packages/web/src/managers/spa-metrics-manager/constants.ts
```diff
@@ -22,6 +22,7 @@ export const BROWSER_NAVIGATION_LOADING_RESOURCE_COUNT_ATTRIBUTE = 'browser.navi
 export const BROWSER_NAVIGATION_LOADING_RESOURCE_URLS_ATTRIBUTE = 'browser.navigation.loading_resource_urls'
 export const BROWSER_NAVIGATION_LONGEST_LOADED_RESOURCE_ATTRIBUTE = 'browser.navigation.longest_loaded_resource'
 export const BROWSER_NAVIGATION_PAGE_COMPLETION_TIME_ATTRIBUTE = 'browser.navigation.page_completion_time'
+export const BROWSER_NAVIGATION_PAGE_SPAN_ID_ATTRIBUTE = 'browser.navigation.page_span_id'
```
Abinet18, line 25: Add browser.navigation.pct_relevant - boolean, page_span_id should be always available

packages/integration-tests/src/tests/user-interaction/spa-metrics.spec.ts
```diff
@@ -116,6 +170,29 @@ test.describe('spa-metrics', () => {
...
+		const afterPctFetchSpan = recordPage.receivedSpans.find(
+			(span) => span.attributes['http.url'] === 'http://localhost:3000/some-data?resource=after-pct-fetch',
+		)
+
+		expectDefined(routeChangeSpan)
+		expectDefined(afterPctFetchSpan)
+		expect(afterPctFetchSpan).toHaveSpanAttribute(BROWSER_NAVIGATION_ATTRIBUTES.pageSpanId, routeChangeSpan.spanId)
+		expect(afterPctFetchSpan).toHaveSpanAttribute(BROWSER_NAVIGATION_ATTRIBUTES.pctRelevant, false)
+	})
+
```
mvirgil, line 195: Could we add an overlapping-navigation regression test here? Start navigation A with a delayed fetch, start navigation B before A’s fetch completes, then assert that the fetch span’s browser.navigation.page_span_id still references A rather than B.

(for both fetch and xhr)

Abinet18: Thanks for bringing this up, 

This requires browser.navigation.page_span_id must be captured when a network request starts, not when it completes.
Previously, Fetch assigned the navigation attributes from its completion callback. A Fetch started during navigation A but completed during navigation B could therefore be associated with B. XHR assigned the attributes when the span was created, but reassigned them on every readystatechange event, allowing the original value to be overwritten after another navigation started.
This change makes both instrumentations capture the active navigation once, when the request span is created:
Fetch assigns the navigation attributes immediately after creating the span.
XHR retains the initial assignment and no longer reassigns them during readystatechange.
The span remains open until the request completes, but its navigation association remains tied to where the request started.
Applying the same behavior to other spans is more difficult. Post-load resource, long-task, and long-animation-frame spans are created retrospectively from PerformanceEntry objects. The browser usually delivers these entries after the activity has completed. By that point, the active navigation may already have changed. Although the span itself can be backdated using entry.startTime or entry.fetchStart, simply reading the current navigation still risks assigning the wrong page ID.
We should consider a timestamp-based navigation lookup as the general solution. SpaMetricsManager could retain a bounded history of navigation start times and span IDs. Instrumentations could then resolve the navigation active at the activity’s actual start time:
```
spaMetricsManager.getNavigationAtTime(entry.fetchStart)
spaMetricsManager.getNavigationAtTime(entry.startTime)
```
This would provide consistent start-time attribution for Fetch, XHR, resources, long tasks, and long animation frames. It is broader than the immediate Fetch/XHR fix, so it would be sensible to handle it as a follow-up change with overlapping-navigation tests for each performance-entry-based span type.

packages/web/src/instrumentations/splunk-post-doc-load-resource-instrumentation.ts
```diff
@@ -117,6 +119,7 @@ export class SplunkPostDocLoadResourceInstrumentation extends InstrumentationBas
 		span.setAttribute('component', MODULE_NAME)
 		span.setAttribute(SemanticAttributes.HTTP_URL, entry.name)
 		span.setAttribute(SemanticAttributes.HTTP_METHOD, 'GET')
+		setBrowserNavigationPageAttributes(span, this.spaMetricsManager)
```
tchoi1131: **Codex review comment — [P2] Resolve the navigation from the activity’s start time**

This assigns the navigation that is current when the `PerformanceObserver` callback runs, while the span itself is backdated to `entry.fetchStart`. These can refer to different navigations.

For example:

```text
0 ms    navigation A starts
100 ms  resource starts
400 ms  navigation B starts
500 ms  resource PerformanceEntry is delivered
```

The resource belongs to A, but this code assigns B’s `page_span_id`.

Fetch and XHR now correctly capture the page ID when the request starts. Could we provide equivalent behavior for PerformanceEntry-based spans by retaining a small, bounded navigation timeline and resolving from `entry.fetchStart` or `entry.startTime`?

```ts
spaMetricsManager.getNavigationAtTime(entry.fetchStart)
```

The same issue applies to post-load resources, long tasks, LoAFs, and document resource spans. Please also add an overlapping-navigation regression test for at least one retrospective span type.

packages/web/src/managers/spa-metrics-manager/navigation-relevance.ts
```diff
@@ -0,0 +1,33 @@
...
+import { BROWSER_NAVIGATION_PAGE_SPAN_ID_ATTRIBUTE, BROWSER_NAVIGATION_PCT_RELEVANT_ATTRIBUTE } from './constants'
+import { type SpaMetricsManager } from './spa-metrics-manager'
+
+export function setBrowserNavigationPageAttributes(span: Span, spaMetricsManager: SpaMetricsManager | undefined): void {
+	const navigationSpanId = spaMetricsManager?.getCurrentNavigationSpanId()
+	if (navigationSpanId) {
+		span.setAttribute(BROWSER_NAVIGATION_PAGE_SPAN_ID_ATTRIBUTE, navigationSpanId)
+		span.setAttribute(
+			BROWSER_NAVIGATION_PCT_RELEVANT_ATTRIBUTE,
+			spaMetricsManager?.isCurrentNavigationPctRelevant() ?? false,
```
tchoi1131: **Codex review comment — [P2] Derive `pct_relevant` from the resource tracking decision**

This flag only indicates that the current navigation’s PCT window is open. It does not establish that this specific resource affected the PCT calculation.

For example, suppose network monitoring is disabled for a route:

```ts
{
  match: '/settings',
  monitors: ['media', 'performance'],
}
```

A fetch during that navigation receives `pct_relevant=true`, even though the manager rejects it and PCT does not wait for it. The same false positive occurs for ignored URLs, the beacon endpoint, and resources beyond `maxResourcesToWatch`.

Could the manager preserve and return the outcome of its existing resource-admission checks instead?

```json
{
  "pageSpanId": "settings-navigation",
  "pctRelevant": false
}
```

If the intended meaning is only “this span started while the PCT window was open,” then the implementation is consistent, but the attribute and PR description should use that weaker meaning—for example, `pct_window_open`. The current description says these spans “contribute to page completion timing.”

packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts
```diff
@@ -118,6 +118,10 @@ export interface SpaMetricsManagerConfig extends SpaMetricsManagerConfigValues {
 export class SpaMetricsManager {
```
tchoi1131, line 149: **Codex review comment — [P3] Consider a name that covers document and soft navigations**

This manager is now used by document-load, fetch, XHR, resource, long-task, and LoAF instrumentations. Several of those apply to traditional document loads as well as SPA route changes, so names such as this are confusing:

```ts
SplunkDocumentLoadInstrumentation(
  /* ... */
  spaMetricsManager,
)
```

Could we rename the internal class to `BrowserNavigationMetricsManager`? That better covers initial document navigations, SPA route changes, PCT calculation, and the new page-span attribution.

The public `spaMetrics` configuration can remain unchanged for compatibility. I don’t think this requires introducing a new interface or extracting another class while attribution remains limited to the current page ID and PCT state.

packages/web/src/instrumentations/splunk-post-doc-load-resource-instrumentation.ts
Abinet18: done

packages/web/src/managers/spa-metrics-manager/navigation-relevance.ts
Abinet18: Tried to handle this by adding admission check for spans within active pct window

packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts
Abinet18: I would handle this in separate PR

packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts
```diff
@@ -337,6 +441,36 @@ export class SpaMetricsManager {
...
+		const normalizedUrl = this.normalizeResourceUrl(activity.url)
+		const matchingDecisions = Array.from(this.resourceAdmissionDecisions.values()).filter(
+			(decision) =>
+				!decision.consumed &&
+				activity.monitorTypes.includes(decision.monitorType) &&
+				decision.url === normalizedUrl &&
+				Math.abs(decision.startTime - startTime) <= RESOURCE_ADMISSION_START_TIME_TOLERANCE,
+		)
+
+		for (const decision of matchingDecisions) {
```
tchoi1131, line 478: **[P2] Avoid consuming every fuzzy admission match**

Review comment from Codex: When same-URL requests start close together, one span can match and consume admission decisions belonging to multiple requests. For example, decisions at 1000 ms and 1050 ms both match a span starting at 1002 ms under the ±100 ms tolerance. This loop consumes both, so the second span later finds no decision and receives `pct_relevant=false`.

Could we propagate the monitor's existing `event.id` to the instrumentation and use the exact-ID lookup? That would avoid fuzzy matching and remove the need to consume decisions. If sharing the ID is too large a change for this PR, could we instead claim only the closest unconsumed decision per applicable monitor type? A regression test with two concurrent requests to the same URL would cover this case.

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
```diff
@@ -132,6 +134,12 @@ export class SplunkDocumentLoadInstrumentation extends DocumentLoadInstrumentati
 			}
 
 			if (span && exposedSpan.name !== AttributeNames.DOCUMENT_LOAD) {
+				const startTime = (entries as unknown as Record<string, unknown>)[performanceName]
+
+				if (typeof startTime === 'number') {
+					setBrowserNavigationPageAttributes(span, this.spaMetricsManager, startTime, { type: 'document' })
```
tchoi1131: **[P2] Apply resource admission to document resource spans**

Review comment from Codex: This branch handles both `documentFetch` and individual `resourceFetch` spans as `{ type: 'document' }`. The document activity path bypasses `evaluateResourceAdmission()`, so a resource rejected because its monitor is disabled, its URL is ignored, or `maxResourcesToWatch` was reached can still receive `pct_relevant=true`.

Could we keep `documentFetch` as `{ type: 'document' }`, but pass `resourceFetch` as `{ type: 'resource', url, monitorTypes }` so it uses the admission result? For example, an ignored script resource should retain its `page_span_id` but have `pct_relevant=false`.

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
```diff
@@ -132,6 +134,12 @@ export class SplunkDocumentLoadInstrumentation extends DocumentLoadInstrumentati
 			}
 
 			if (span && exposedSpan.name !== AttributeNames.DOCUMENT_LOAD) {
+				const startTime = (entries as unknown as Record<string, unknown>)[performanceName]
```
tchoi1131: **[P2] Resolve the navigation using the resource start time**

Review comment from Codex: The upstream document-load instrumentation calls `_endSpan()` with `performanceName = RESPONSE_END`, so `entries[performanceName]` is the resource completion time rather than its start time. If a resource starts under navigation A and finishes after navigation B begins, this can associate the span with B.

Could we use `entries[PTN.FETCH_START]` for the navigation lookup? For example, a resource starting at 1200 ms under A, followed by navigation B at 1500 ms and `responseEnd` at 1800 ms, should still use A's `page_span_id`. A regression test covering a resource that crosses a navigation boundary would help protect this behavior.

packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts
Abinet18: updated

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
Abinet18: updated as per comment

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
Abinet18: updated

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
```diff
@@ -132,6 +135,28 @@ export class SplunkDocumentLoadInstrumentation extends DocumentLoadInstrumentati
 			}
 
 			if (span && exposedSpan.name !== AttributeNames.DOCUMENT_LOAD) {
+				const isResourceFetch = exposedSpan.name === AttributeNames.RESOURCE_FETCH
+				const startTime = (entries as unknown as Record<string, unknown>)[
+					isResourceFetch ? PTN.FETCH_START : performanceName
```
tchoi1131: **[P2] Use `fetchStart` when resolving `documentFetch`**

Review comment from Codex: This update fixes `resourceFetch` by selecting `PTN.FETCH_START`, but `documentFetch` still selects `entries[performanceName]`. For `documentFetch`, `performanceName` is normally `PTN.RESPONSE_END`, so this value is passed to `SpaMetricsManager.getNavigationPageAttributes()`, which resolves the page span based on the document's completion time.

For example, if a streamed document starts fetching at `0 ms`, application code starts a route navigation at `500 ms`, and the document response ends at `800 ms`, the `documentFetch` span can be associated with the route navigation even though it began under the document navigation.

Could we use `PTN.FETCH_START` for both `documentFetch` and `resourceFetch`?

```ts
const startTime = (entries as unknown as Record<string, unknown>)[PTN.FETCH_START]
```

`resourceFetch` would still use `{ type: 'resource', ... }`, while `documentFetch` would continue using `{ type: 'document' }`. A regression test where a streamed document response crosses a route-navigation boundary would help protect this behavior.

Code references:
- [Current timestamp selection](https://github.com/signalfx/splunk-otel-js-web/blob/85495dcadba22ed9340ab716850498b854d04c51/packages/web/src/instrumentations/splunk-document-load-instrumentation.ts#L137-L155)
- [Navigation lookup](https://github.com/signalfx/splunk-otel-js-web/blob/85495dcadba22ed9340ab716850498b854d04c51/packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts#L224-L246)

mvirgil (approved):

tchoi1131 (approved):
Review from Codex: The previously identified navigation-attribution issue is resolved, and the focused streaming-navigation regression test passes in both Chrome and Firefox. No remaining blocking findings.

PR #1879, opened 2026-07-22
feat: Introduce spans for selector blocked elements 

shumone:
# Description

This PR introduces a new instrumentation, `blockingElement`. The purpose of the instrumentation is to get visibility on customer chosen browser selectors (like loading-spinner, various css elements, svg anims, etc), that may contribute to loading times that are blocking (see `loading-element-monitor` in `spa-metrics-monitor` for a similar implementation). 

The instrumentation follows closely w/ the `loading-element-monitor`, with a `mutationObserver` for tracking elements that may have loaded. It differs slightly in that it tracks all elements for a given selector and emits a span for each element, instead of grouping them together like in the monitor. 

Configs are taken from `blocking-element` first, then from `spaMetrics` next if `monitors` contains `elements` and has `blockingSelectors` in it. 

An example config for `blocking-element`: 

```
SplunkRum.init({
    ...
    instrumentations: {
      blockingElement: {
        enabled: true,
        selectors: ['.loading-spinner', '[data-loading]'],
      },
    },                              
  })
```
  
 An example config from `spaMetrics`: 
 
```
spaMetrics: {
    monitors: ['media', 'network', 'performance', 'elements'],
    blockingSelectors: ['.loading-spinner'],
    urlOverrides: [
      {
        match: '/checkout/',
        monitors: ['elements'],
        blockingSelectors: ['[data-checkout-loading]'],
      },                            
    ],
  }
```

There were several design considerations/todos for future tickets; one is the config `urlOverrides` from `spaMetrics`. This MR hasn't added the navigation route-change handling yet, which means when we inherit it, it'll only work for the current base page. 

## Type of change

# How has this been tested?

- Manual testing (TODO)
- Added unit tests
- Added integration tests


packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
```diff
@@ -0,0 +1,136 @@
...
+	get openCount(): number {
+		let count = 0
+		for (const elements of this.spansBySelector.values()) {
+			count += elements.size
+		}
+
+		return count
+	}
+
+	startSpan(selector: string, element: Element, startTimeRelative: number): void {
```
Abinet18: [P1] A spinner that remains visible never exports a span.
Spans end only when the element disappears or instrumentation is disabled. There is no timeout or pagehide termination path. The integration test explicitly accepts that the open-ended spinner produces no telemetry in [blocking-element-span.spec.ts (line 63)]
This loses the most important failure case: a spinner that stays indefinitely. Add a configurable maximum duration and finish with completion="timeout", plus interrupt open spans on pagehide.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
```diff
@@ -0,0 +1,162 @@
...
+	private clearScheduledScan(): void {
+		if (this.scanTimeoutId === undefined) {
+			return
+		}
+
+		clearTimeout(this.scanTimeoutId)
+		this.scanTimeoutId = undefined
+	}
+
+	private getVisibleElements(selector: string): Element[] {
```
Abinet18: [P2] Element monitoring is substantially duplicated.
[splunk-blocking-element-instrumentation.ts (line 81)] duplicates from LoadingElementMonitor:
Visibility calculation
Invalid-selector handling
MutationObserver setup
Scan batching
Selector scanning and reconciliation
Compare [loading-element-monitor.ts (line 154)] When fallback configuration enables both, the browser runs two document-wide observers and two sets of querySelectorAll calls.
A shared ElementVisibilityObserver should emit per-element visible/hidden events. The PCT monitor can aggregate these by selector, while the span instrumentation creates individual spans.

packages/web/src/instrumentations/blocking-element/support.ts
```diff
@@ -0,0 +1,89 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import type { SpaMetricsOptions, SplunkOtelWebConfig } from '../../types'
+
+function isUrlOverrideMatch(match: string | RegExp, url: string): boolean {
```
Abinet18, line 23: [P2] SPA configuration resolution is also duplicated and becomes stale.
[support.ts (line 21)] recreates URL override matching and partial inheritance already owned by SpaMetricsManager. It resolves only once during enable().
Therefore:
Starting on a route where the override disables element monitoring can disable spans for the entire session.
Starting on an enabled route continues using those selectors after navigating to a route where they should be disabled.
Later URL override selector changes are ignored.
The README documents this limitation, but it weakens the claimed fallback to spaMetrics.urlOverrides. Configuration resolution should be shared with SpaMetricsManager and refreshed on navigation.

packages/web/src/instrumentations/blocking-element/support.ts
```diff
@@ -0,0 +1,89 @@
...
+	return {
+		...override,
+		blockingSelectors: override.blockingSelectors ?? spaMetrics.blockingSelectors,
+		monitors: override.monitors ?? spaMetrics.monitors,
+	}
+}
+
+// Reads the raw instrumentations.blockingElement value, not this._config/getConfig() —
+// getPluginConfig's pluginDefaults force enabled to false on every instrumentation's processed config.
+export function isBlockingElementInstrumentationEnabled(otelConfig: SplunkOtelWebConfig): boolean {
```
Abinet18, line 50: [P2] Supplying selectors alone does not enable the instrumentation.
With:
instrumentations: {
  blockingElement: {
    selectors: ['.spinner'],
  },
}
the instrumentation remains disabled unless the current SPA metrics configuration also enables the elements monitor. This follows [support.ts (line 48)] but differs from the normal instrumentation convention that supplying a configuration object enables it.
Either object presence should enable it, or the README must clearly require enabled: true.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
```diff
@@ -0,0 +1,136 @@
...
+		if (this.has(selector, element)) {
+			return
+		}
+
```
Abinet18, line 128: [P2] No span-volume safety limit exists.
Every visible selector/element pair opens a root span. A broad or mistaken selector such as div, combined with frequent DOM churn, can create a large number of spans and retain many DOM references. openCount mentions a future cap, but no cap is implemented.
Add at least:
maxSpansPerPage or a rolling-window limit
Maximum simultaneously open spans
A timeout for each open span

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/**
+ * Creates one span per (selector, element) pair matching a configured blocking selector.
+ * Kept fully independent of LoadingElementMonitor's own PCT tracking: this class only reacts
+ * to element-level start/complete/interrupt calls it is given and never reads or mutates PCT
+ * state.
+ */
+export class ElementSpanTracker {
+	private readonly spansBySelector = new Map<string, Map<Element, Span>>()
```
Abinet18: [P2] Overlapping selectors create duplicate spans for one element.
 An element matching .spinner and #checkout-spinner creates two spans.
That conflicts with the README’s “one span per DOM element” wording. Decide explicitly:
One span per element: use Map<Element, { span, selectors: Set<string> }>
One span per selector/element pair: retain the current structure, document it, and test overlapping selectors

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
```diff
@@ -0,0 +1,136 @@
...
+	startSpan(selector: string, element: Element, startTimeRelative: number): void {
+		if (this.has(selector, element)) {
+			return
+		}
+
+		const span = this.tracer.startSpan(BLOCKING_ELEMENT_SPAN_NAME, {
+			root: true,
+			startTime: startTimeRelative,
+		})
+		span.setAttribute('component', BLOCKING_ELEMENT_SPAN_NAME)
```
Abinet18: [P2] The component attribute is inconsistent with other agent spans.
It is set to the span name, blockingElement, Other instrumentations use their instrumentation/module name, such as splunk-longtask or splunk-loaf.
It should likely be splunk-blocking-element, especially if backend or UI classification uses component.

Abinet18:
**Suggested tests to add**

Stuck-spinner timeout/page-exit export
Explicit { selectors: [...] } configuration without enabled
URL override changes after route navigation
Overlapping selectors
Span-volume cap
Hide/show and stop-matching transitions end-to-end
Duplicate enable() calls
Correct component value

Abinet18 (changes requested):

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
shumone: @Abinet18 Added a guard in constants, set to 1000. The guard caps against the maximum number of open spans allowed to be tracked (increases on startSpan, decreases on endSpan).

Do you think thats more appropriate, or a cap on the number of spans allowed to be created across the instrumentation's lifespan?

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
shumone: Whoops, this should be using `BLOCKING_ELEMENT_MODULE_NAME`. Replaced that with the instrumentation module name.

packages/web/src/instrumentations/blocking-element/support.ts
shumone: Reworked this. The config for our instrumentation now only has enable/disable (disabled by default if not specified, but becomes enabled if spaMetrics contains element in monitors + has blockedSelectors). We only use the config from spaMetrics now. 

The note about urlOverrides is accurate; I want to work on this in a separate ticket as the PR is getting larger and larger. In that ticket, I will include navigation routechange to the instrumentation, and then update the config based on urlOverrides then as required.

packages/web/src/instrumentations/blocking-element/support.ts
shumone: Gotcha, will change the readme to default enabled = true. 
Additionally, we're not including any options/configs inside `blockingElement` anymore. It only has enabled or disabled now. The only way to disable the instrumentation will be if its explicitly disabled, otherwise it will try to use spaMetrics as usual.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
shumone: I think this is valid and I was considering this, but will take some time to refactor.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
shumone: Resolved; chose to use Map<Element, {span, selectors}> instead.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
Abinet18: This is fine too, but 1000 may be too much for concurrent spans , may be we can leave it as is with a todo to revisit the value

packages/web/src/instrumentations/blocking-element/support.ts
Abinet18: Ok, let us have the followup tickets created.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
Abinet18: ok, you can create the followup ticket to handle this

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
```diff
@@ -0,0 +1,162 @@
...
+			this.interruptSpan(element)
+		}
+	}
+
+	/**
+	 * Interrupts every open span whose element currently matches the given selector. No caller in
+	 * this instrumentation's v1 scope — kept as the seam a future per-route selector re-resolution
+	 * ticket would call for each selector that falls out of a newly resolved list.
+	 */
+	interruptAllForSelector(selector: string): void {
```
Abinet18: This is not used except in the tests, Can it be removed

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
```diff
@@ -0,0 +1,123 @@
...
+	private readonly handleVisibilityChange = (event: ElementVisibilityChangeEvent): void => {
+		const elementSpanTracker = this.elementSpanTracker
+		if (!elementSpanTracker) {
+			return
+		}
+
+		const { element, visible } = event
+		const now = performance.now()
+
+		if (!visible) {
```
Abinet18, line 151: [P1] Multi-selector elements can complete prematurely.
If an element matches .spinner and [data-loading], then stops matching only [data-loading], the observer emits visible:false for that selector. The instrumentation ends the span even though the element remains visible as .spinner. May be It should maintain the currently active selector set per element and end only when that set becomes empty. The comment here 'stopping matching one selector while still matching another can't happen' doesn't feel right

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
shumone: Added a timeout config, with a `maxElementSpanDuration` config set to a default of 3 minutes. 

Added the pagehide handler, which adds the interrupt path for those open spans.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
shumone: Added a todo comment. Maybe can see what a sensible value is when testing on lab0.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
shumone: Expecting to use this for urlOverride in separate ticket, but can remove it now since its not used.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
shumone: Yea, I think you're right. Changed to be if the set is empty then it ends.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
```diff
@@ -0,0 +1,150 @@
...
+
+		const { element, selector, visible } = event
+		const now = performance.now()
+
+		if (!visible) {
+			// The observer's events are per (selector, element) pair, not per element — an element
+			// matching two configured selectors can drop out of one while still matching the other,
+			// producing a visible:false for only the dropped selector. Only complete the span once
+			// the element's active-selector set is empty, i.e. it no longer matches anything configured.
+			const activeSelectors = this.activeSelectorsByElement.get(element)
```
Abinet18, line 156: [P2] Switching directly between selectors can still split one continuous spinner into two spans. The first visible:false event may empty the set before the observer delivers visible:true for the replacement selector.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
```diff
@@ -0,0 +1,150 @@
...
+import { BLOCKING_ELEMENT_MODULE_NAME } from './blocking-element/constants'
+import { ElementSpanTracker } from './blocking-element/element-span-tracker'
+import {
+	isBlockingElementInstrumentationEnabled,
+	resolveBlockingElementSelectors,
+	resolveMaxElementSpanDuration,
+} from './blocking-element/support'
+
+// once: true means the browser removes this listener for us after it fires; no re-entry guard needed.
+const PAGEHIDE_LISTENER_OPTIONS: AddEventListenerOptions = { capture: true, once: true }
```
Abinet18: [P2] The pagehide listener uses once: true at. After a BFCache restoration, newly created spans will not be interrupted on the next pagehide. Remove once, and consider resyncing existing visible elements on pageshow.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
```diff
@@ -0,0 +1,162 @@
...
+			root: true,
+			startTime: startTimeRelative,
+		})
+		span.setAttribute('component', BLOCKING_ELEMENT_MODULE_NAME)
+		span.setAttribute(BROWSER_ELEMENT_ID_ATTRIBUTE, element.id)
+		span.setAttribute(BROWSER_ELEMENT_CLASS_ATTRIBUTE, getElementClass(element))
+		span.setAttribute(BROWSER_ELEMENT_TAG_ATTRIBUTE, element.tagName)
+		span.setAttribute(BROWSER_ELEMENT_XPATH_ATTRIBUTE, getElementXPath(element, true))
+
+		const timeoutId = setTimeout(() => {
```
Abinet18, line 139: [P2] After timeout removes a span , the instrumentation does not mark the still-visible element as timed out. Gaining another selector can start a second span for the same uninterrupted visibility episode.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
shumone: Proposed fix: Change `scanSelectors` to notify visible events first (so the replacement selector), then the disappeared event (the dropped selector). When the disappeared event arrives, the consumer should not complete the span yet, because the replacement selector is still considered an active selector in the instrumentation.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
shumone: Included pageshow handler (only for event.persisted = true), and resyncs. Also removed the once:true since that is no longer relevant.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
shumone: Added a onSpanTimeout callback in the instrumentation to track timed-out elements. A visible element doesn't get to startSpan if its in that set of tracked timedout elements. If it becomes non-visible, it is removed from that set.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
```diff
@@ -0,0 +1,185 @@
...
+		this.elementSpanTracker?.interruptAll()
+		this.elementSpanTracker = undefined
+		this.selectors = []
+		this.activeSelectorsByElement.clear()
+		this.timedOutElements.clear()
+		this.elementVisibilityObserver.unwatch(this.consumerId)
+	}
+
+	enable(): void {
+		if (!isBlockingElementInstrumentationEnabled(this.otelConfig)) {
```
Abinet18, line 94: [P2] Calling enable() twice replaces elementSpanTracker without ending the spans owned by the previous tracker. The observer’s watch() call then no-ops because the same consumer already watches those selectors. When an existing element disappears, completion is sent to the new empty tracker, leaving the original span open until timeout. Add an enabled guard or make repeated enable() a no-op in

shumone: Addressed by adding a new guard on enable (checks whether this.elementSpanTracker is present. The tracker is initialized in enable, torn down in disable.)

Abinet18:
**Two more issues reported by codex, I think these could be handled on the followup,** 

[P1] Interrupt spans before the last reliable flush. The only interruption occurs on pagehide at [splunk-blocking-element-instrumentation.ts:121), while the existing hidden-state flush happens at [index.ts (line 751)]. pagehide is not guaranteed, and may occur after the hidden-state flush, leaving newly ended spans buffered and lost. The hidden transition is considered the last reliable opportunity to send analytics. ([Chrome Page Lifecycle guidance](https://developer.chrome.com/docs/web-platform/page-lifecycle-api#developer-recommendations-for-each-state)) Interrupt before that flush, then resync visible elements when the page resumes.

[P2] Re-resolve selectors on SPA navigation. Selectors and derived enablement are resolved only during enable() at [splunk-blocking-element-instrumentation.ts (line 98)]. Meanwhile, PCT refreshes spaMetrics.urlOverrides per navigation. After navigating, blocking spans can continue using the previous route’s selectors or miss the new route’s selectors entirely. The instrumentation should refresh alongside SpaMetricsManager.

Abinet18 (approved):

PR #1883, opened 2026-07-28
feat: attribute spans to navigation operations

mvirgil:
## Summary

- add `browser.navigation.operation` to spans that contribute to browser metrics
- use `documentLoad` and `routeChange` values from the active SPA navigation
- cover document-load resources, route-change errors, page spans, and processor behavior

## Screenshots

image
image

## AI Attestation

This pull request was authored with OpenAI Codex. Codex implemented the code and tests, performed a local code review, and assisted with manual validation. The human author reviewed and directed the work.

### DIFF
--- a/packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts
+++ b/packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts
@@ -25,6 +25,7 @@ import type { Monitor, MonitorConfig } from './monitors/monitor'
 import { truncateString } from '../../utils/text'
 import {
 	BROWSER_NAVIGATION_DETECTED_RESOURCE_COUNT_ATTRIBUTE,
+	BROWSER_NAVIGATION_DOCUMENT_LOAD_OPERATION,
 	BROWSER_NAVIGATION_LAST_LOADED_RESOURCES_ATTRIBUTE,
 	BROWSER_NAVIGATION_LOADING_RESOURCE_COUNT_ATTRIBUTE,
 	BROWSER_NAVIGATION_LOADING_RESOURCE_URLS_ATTRIBUTE,
@@ -102,11 +103,13 @@ type DroppedLoadingResources = {
 }
 
 type WaitForPageLoadConfig = {
+	operation?: string
 	span?: Span
 	startTime: number
 }
 
 type NavigationHistoryEntry = {
+	operation: string
 	pctEndTime?: number
 	spanId: string
 	startTime: number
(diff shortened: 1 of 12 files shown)


packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts
```diff
@@ -247,7 +255,8 @@ export class SpaMetricsManager {
 	}
 
 	setCurrentNavigationSpan(span: Span, startTime: number): void {
-		this.navigationHistory.push({ spanId: span.spanContext().spanId, startTime })
+		const operation = isSpanWithName(span) ? span.name : BROWSER_NAVIGATION_DOCUMENT_LOAD_OPERATION
```
Abinet18: Codex review: 
[P2] Route operations are inferred from a non-standard span.name property . The OpenTelemetry Span API does not guarantee this property; sampled-out spans are NonRecordingSpan objects without it. A sampled-out routeChange therefore gets stored as documentLoad, causing later sampled spans to be misattributed. Pass the operation explicitly when registering the navigation. `setCurrentNavigationSpan(span, startTime, 'routeChange')`

mvirgil: Done

packages/web/src/span-processors/span-attributes-processor.ts
```diff
@@ -65,6 +71,18 @@ export class SpanAttributesProcessor implements SpanProcessor {
 	}
 
 	onStart(span: Span): void {
+		if (this.spaMetricsManager) {
+			const startTime = hrTimeToMilliseconds(span.startTime) - performance.timeOrigin
+			// Navigation spans reach processors before instrumentation registers them with SpaMetricsManager,
+			// so use their runtime names instead of looking up the previous navigation.
+			const operation =
```
Abinet18: [P2] Set browser.navigation.operation explicitly at the instrumentation level instead of deriving it from span.name in SpanAttributesProcessor.
The span processor handles every span, while span names are neither unique nor guaranteed identifiers. Application spans can also be named routeChange or documentLoad, instrumentations may rename spans, and sampled-out spans may not expose a name at all. This can cause a span to receive the wrong navigation operation.
The document-load and user-interaction instrumentations already know the operation they are creating, so they should supply it explicitly when starting/registering the navigation span. The processor should use the manager’s navigation history only for other spans. This removes the dependency on SDK-specific span properties, span naming conventions, sampling decisions, and navigation-registration timing.
Navigation instrumentations can set the operation when creating the span:
```
tracer.startSpan('routeChange', {
  startTime,
  attributes: {
    'browser.navigation.operation': 'routeChange',
  },
})
```
Then the processor only fills the attribute when it is absent:
```
if (span.attributes[BROWSER_NAVIGATION_OPERATION_ATTRIBUTE] === undefined) {
  const startTime = hrTimeToMilliseconds(span.startTime) - performance.timeOrigin
  span.setAttribute(
    BROWSER_NAVIGATION_OPERATION_ATTRIBUTE,
    spaMetricsManager.getNavigationOperation(startTime),
  )
}
```

mvirgil: Done

Abinet18 (approved):

PR #1884, opened 2026-07-31
feat: add page span ID to error spans

mvirgil:
# Description

Add the current browser navigation page span ID to automatically captured and manually reported error spans.

The shared error-span creation path now applies browser navigation page attributes, covering both document-load and route-change pages. Integration coverage verifies the page span ID on handled and unhandled errors and confirms route-change errors use the current route's page span.

## Type of change

# How has this been tested?

- Added integration tests for handled and unhandled errors on document-load pages
- Added integration coverage for errors after a route change
- Focused Playwright tests: 6 passed across Chromium, Firefox, and WebKit
- Full unit suite: 1,476 passed, 34 skipped
- Type checks, ESLint, and production build passed
- Manual verification in a test environment:
  - Document-load and route-change pages generated
  - Handled and unhandled errors generated

## AI Attestation

This change was planned, implemented, tested, and documented with assistance from OpenAI Codex. The submitter reviewed the resulting diff and manually verified the behavior in a test environment. No customer data or production credentials are included in the change.

### DIFF
--- a/packages/web/src/instrumentations/splunk-error-instrumentation.ts
+++ b/packages/web/src/instrumentations/splunk-error-instrumentation.ts
@@ -195,7 +195,7 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
 			return
 		}
 
-		const { span, timestamp } = this.startErrorSpan(source)
+		const { now, span } = this.startErrorSpan(source)
 
 		this.attachSpanContext(span, err, spanContext)
 
@@ -208,7 +208,7 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
 		span.setAttribute('error.message', limitLen(msg, MESSAGE_LIMIT))
 		addStackIfUseful(span, err)
 
-		await this.endSpanWithThrottle(span, timestamp)
+		await this.endSpanWithThrottle(span, now)
 	}
 
 	protected async reportErrorEvent(source: string, ev: ErrorEvent, spanContext: SpanContext): Promise<void> {
@@ -225,7 +225,7 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
 			return
 		}
 
-		const { span, timestamp } = this.startErrorSpan(source)
+		const { now, span } = this.startErrorSpan(source)
 		this.attachSpanContext(span, ev, spanContext)
 		span.setAttribute('component', 'error')
 		span.setAttribute('error.type', ev.type)
@@ -256,7 +256,7 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
 			}
 		}
 
-		await this.endSpanWithThrottle(span, timestamp)
+		await this.endSpanWithThrottle(span, now)
 	}
 
 	protected async reportString(
@@ -269,7 +269,7 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
 			return
 		}
 
-		const { span, timestamp } = this.startErrorSpan(source)
+		const { now, span } = this.startErrorSpan(source)
 		this.attachSpanContext(span, firstError ?? message, spanContext)
 		span.setAttribute('component', 'error')
 		span.setAttribute('error', true)
@@ -279,7 +279,7 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
 			addStackIfUseful(span, firstError)
 		}
 
-		await this.endSpanWithThrottle(span, timestamp)
+		await this.endSpanWithThrottle(span, now)
 	}
 
 	protected transform(error: InternalErrorLike, spanContext: SpanContext): ErrorWithContext | null {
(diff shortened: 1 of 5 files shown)


packages/web/src/instrumentations/splunk-error-instrumentation.ts
```diff
@@ -398,6 +397,15 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
 		await this.report('onerror', event, {})
 	}
 
+	private startErrorSpan(source: string): { now: number; span: Span } {
+		const navigationStartTime = performance.now()
+		const now = Date.now()
+		const span = this.tracer.startSpan(source, { startTime: now })
+		setBrowserNavigationPageAttributes(span, this.spaMetricsManager, navigationStartTime)
```
Abinet18: [P1] Pre-load errors are not associated with documentLoad. The lookup in [splunk-error-instrumentation.ts (line 404)] runs immediately, but the document navigation is registered only when OpenTelemetry creates the span after window.load in [splunk-document-load-instrumentation.ts (line 111)]. Errors occurring during initial page execution therefore get no page-span ID and are never backfilled.

packages/web/src/instrumentations/splunk-error-instrumentation.ts
```diff
@@ -398,6 +397,15 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
 		await this.report('onerror', event, {})
 	}
 
+	private startErrorSpan(source: string): { now: number; span: Span } {
```
Abinet18, line 446: [P2] Derive both timestamps from one clock sample
performance.now() and Date.now() represent the same error occurrence but use different clock domains. The page-span lookup uses performance.now(), while SpanAttributesProcessor derives the navigation operation from the span’s Date.now()-based start time. A wall-clock adjustment could make these resolve to different navigations.
Capture const errorStartTime = performance.now() once, use it for the navigation lookup, and derive the span timestamp as performance.timeOrigin + errorStartTime. Return that same epoch timestamp for span.end(). Also, navigationStartTime should be renamed because it represents the error time, not the navigation start.

Abinet18 (commented):
Getting the span id for the pre-load errors is tricky, we can consider queuing them without ending them until the doc load span is created but that could result in missed spans if sth happens in the middle ?

packages/web/src/instrumentations/splunk-error-instrumentation.ts
mvirgil: Addressed in 92dc69f (`fix: associate pre-load errors with document load`). Pre-load error spans are now held until the document-load span is available, with a `pagehide` fallback, while throttling still occurs when the error span is created.

packages/web/src/instrumentations/splunk-error-instrumentation.ts
mvirgil: Addressed in b45638e (`fix: derive error timestamps from one clock`). The span epoch timestamp is now derived from the same `performance.now()` sample used for navigation lookup, and the variable is renamed to `errorStartTime`.

packages/web/src/instrumentations/splunk-error-instrumentation.ts
```diff
@@ -110,32 +111,46 @@ export type SplunkErrorInstrumentationConfig = InstrumentationConfig & {
...
+		window.removeEventListener('pagehide', this.pageHideListener)
 		document.documentElement.removeEventListener('error', this.documentErrorListener, { capture: true })
+		this.otelConfig.spanEmitter?.removeEventListener('document-load:start', this.documentLoadStartListener)
+		this.flushPendingPreLoadSpans()
 		clearInterval(this.clearingIntervalId)
 		this.throttleMap.clear()
 	}
 
 	enable(): void {
+		this.acceptingPendingPreLoadSpans = document.readyState !== 'complete'
```
Abinet18, line 147: [P2] Keep buffering while deferred document-load collection is pending.
When the agent initializes after window.load, readyState is already complete, but OpenTelemetry schedules document-load collection with setTimeout. An error reported immediately after SplunkRum.init() therefore sees no navigation and is not buffered, so it misses browser.navigation.page_span_id. Buffer until document-load:start or a definite collection-complete fallback, and add a late-initialization test.

packages/web/src/instrumentations/splunk-error-instrumentation.ts
```diff
@@ -398,6 +430,37 @@ export class SplunkErrorInstrumentation extends InstrumentationBase {
...
 
+	private flushPendingPreLoadSpans = () => {
+		for (const [span, { endTime, startTime }] of this.pendingPreLoadSpans) {
+			setBrowserNavigationPageAttributes(span, this.spaMetricsManager, startTime)
+			this.pendingPreLoadSpans.delete(span)
+			span.end(endTime)
+		}
+	}
+
+	private pageHideListener = () => {
```
Abinet18, line 441: [P2] Export interrupted pre-load errors after ending them
Pending errors are ended on pagehide, but the agent force-flushes on the earlier visibilitychange: hidden event. For a page interrupted before document load, the flush can run while these spans are still pending; they are ended afterward without another forced export and may be lost. End them before the hidden-page flush or explicitly flush after pagehide.

Abinet18:
I see that this is not specific to errors; other instrumentations may also create spans before the document-load span ID is available. We have two options:
1. Centralize pending spans in SpaMetricsManager, then attach page attributes and end them once the page span is available.
2. Create and register the actual document-load span earlier, populate its attributes later, and end it after PCT completion or interruption.
Second option is better because it provides consistent page association without adding buffering, interruption, and flush complexity to every instrumentation, but we can discuss it

PR #1888, opened 2026-08-05
feat: Rename spa to navigation metrics in the codebase

shumone:
# Description

This MR renames SPA (single page application) to Navigation instead throughout the codebase. 

Includes name changes to the classes that have Spa in them (spa-metrics-manager -> navigation-metrics-manager). 
Includes a new file `resolve-config.ts` to resolve the new `navigationMetrics` config with the now deprecated `spaMetrics` config (kept to avoid disruption with customers).  

## Type of change

- Documentation update

# How has this been tested?

- Manual testing
- Added unit tests


packages/web/src/managers/navigation-metrics-manager/quiet-period-awaiter.ts
Abinet18: nit: rename spa.maxPageLoadWaitTime

packages/web/src/index.ts
```diff
@@ -654,19 +653,20 @@ export const SplunkRum: SplunkOtelWebType = {
...
-			// Shared by SpaMetricsManager (LoadingElementMonitor) and SplunkBlockingElementInstrumentation
+			// Shared by NavigationMetricsManager (LoadingElementMonitor) and SplunkBlockingElementInstrumentation
 			// so both watch the DOM through one MutationObserver instead of two independent ones.
 			const elementVisibilityObserver = new ElementVisibilityObserver()
 
-			const spaMetricsManager =
-				processedOptions.spaMetrics === false
+			const navigationMetricsConfig = resolveNavigationMetricsConfig(processedOptions)
+			const navigationMetricsManager =
+				navigationMetricsConfig === false
```
Abinet18, line 663: Should we Preserve navigation context when navigation metrics are disabled
navigationMetrics: false currently prevents NavigationMetricsManager from being created. The manager now owns general navigation state used for browser.navigation.operation and browser.navigation.page_span_id, so disabling PCT also removes unrelated navigation association. We can instantiate the navigation manager unconditionally and gate only its resource monitors and PCT calculation. Route changes should still be registered and ended immediately when metrics are disabled. 

This can be handled on a followup task if that is better.

packages/web/src/managers/navigation-metrics-manager/quiet-period-awaiter.ts
shumone: Fixed this + 2 other test cases using spa.maxPageLoadWaitTime.

packages/web/src/index.ts
shumone: Lets do this in a separate ticket, since this seems to also have been in place before this ticket was created.

Abinet18 (approved):

PR #1892, opened 2026-08-13
fix: assign page span ID to early spans

Abinet18:
**Description**

Fixes missing browser.navigation.page_span_id attributes on spans created before the documentLoad span is available.
The pageLoad span is now created and registered as the current navigation as soon as navigation timing becomes available. Its start time remains aligned with navigation.fetchStart, documentLoad, and documentFetch. A buffered PerformanceObserver provides a fallback when the navigation entry is not immediately available.
Early fetches and other spans can therefore be associated with the correct page-load span without waiting for document load processing.

**Fixes:** [DEMRUM-6302](https://splunk.atlassian.net/browse/DEMRUM-6302)

**Type of change**

Bug fix (non-breaking change which fixes an issue)

**How has this been tested?**

Added unit tests
Added integration tests
Validation completed:
Lint passed
Package builds and TypeScript compilation passed
1,749 unit tests passed
789 integration tests passed
Verified an actual fetch started while document.readyState === "loading" receives the correct pageLoad span ID across Chromium, Firefox, WebKit, Edge, and Chrome

**AI attestation**
AI assistance was used for implementation, test creation, and validation. The resulting changes were reviewed against existing repository patterns, and the relevant lint, build, unit, and integration checks were run before commit.


packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
```diff
@@ -290,4 +298,76 @@ export class SplunkDocumentLoadInstrumentation extends DocumentLoadInstrumentati
...
+		this.navigationStartTimeMillis ??= hrTimeToMilliseconds(timeInputToHrTime(fetchStart))
+
+		if (this.pageLoadSpan || !this.documentLoadMetricsPromise || !this.otelConfig.experimental) {
+			return
+		}
+
+		this.navigationTimingObserver?.disconnect()
+		this.navigationTimingObserver = undefined
+
+		this.pageLoadSpan = this.tracer.startSpan(PAGE_LOAD_SPAN_NAME, {
```
tchoi1131, line 326: **Codex review comment**

Could we clarify the intended trace topology for `pageLoad` now that it starts earlier from `setTracerProvider()`?

Previously, it was created from the document-load path while the upstream instrumentation’s extracted `<meta name="traceparent">` context was active. In the new path, `startSpan()` runs before that extraction and implicitly inherits `context.active()`.

I think we should make one of these behaviors explicit:

1. If `pageLoad` should continue the server trace, extract the document’s `traceparent` and pass that context as the third argument to `startSpan()`.
2. If `pageLoad` should always be a separate root trace, pass `api.ROOT_CONTEXT` explicitly so initialization inside an unrelated active span cannot accidentally parent it.

The new `browser.navigation.page_span_id` attribute associates activity with the page load, but it does not determine the OpenTelemetry trace or parent relationship.

Could we confirm which topology is intended and add a test asserting the resulting `traceId` and `parentSpanId`?

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
```diff
@@ -290,4 +298,76 @@ export class SplunkDocumentLoadInstrumentation extends DocumentLoadInstrumentati
...
+		if (
+			!this.documentLoadMetricsPromise ||
+			!this.otelConfig.experimental ||
+			this.pageLoadSpan ||
+			this.navigationTimingObserver
+		) {
+			return
+		}
+
+		const navigationEntry = performance.getEntriesByType('navigation')[0] as PerformanceNavigationTiming | undefined
```
shumone, line 349: Claude [P2] — Early spans get no `page_span_id` at all when the async `PerformanceObserver` fallback fires

If `performance.getEntriesByType('navigation')[0]` isn't populated synchronously at `setTracerProvider()` time, `pageLoadSpan` creation is deferred to an async `PerformanceObserver` callback. Any span (e.g. an early `fetch`, per the new `docload-early-fetch.ejs` test) that starts before that callback fires finds `SpaMetricsManager`'s `navigationHistory` still empty — `getNavigationAt` returns `undefined`, `getNavigationPageAttributes` returns `undefined`, and `setBrowserNavigationPageAttributes` silently sets no `browser.navigation.page_span_id`/`pct_relevant` at all. So in browsers/timings where the sync path misses, this regresses from "wrong attribution" (the pre-fix bug) to "no attribution" — for precisely the early-span case this branch targets.

Suggest adding a test that forces the fallback branch (stub `performance.getEntriesByType` to return `[]` on the first call) and asserts an early span still gets attributed correctly. For the fix itself: either buffer/backfill navigation attributes for spans that arrive before `pageLoadSpan` exists, or seed a placeholder navigation-history entry at `setTracerProvider()` time so `getNavigationAt` never returns `undefined` during this window, updating it once `pageLoadSpan` is actually created.

human note: My understanding is this is flagging spans that start during that waiting window in between `setTracerProvider` returning and the callback you create around line 366. I'm not sure if its very likely a span would happen in that time window (since the callback is async), but maybe its a concern?

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
```diff
@@ -290,4 +298,76 @@ export class SplunkDocumentLoadInstrumentation extends DocumentLoadInstrumentati
 			}
 		}
 	}
+
+	disable(): void {
```
shumone, line 302: Claude [P3] — `disable()` doesn't end an already-open `pageLoadSpan`

`disable()` only disconnects `navigationTimingObserver` and calls `super.disable()`, which upstream just removes the `load` listener — neither touches `this.pageLoadSpan`. If the instrumentation is disabled after `pageLoadSpan` was created (via the sync or `PerformanceObserver` path) but before the `load` event fires, the span is never ended or cleared: the only code that calls `pageLoadSpan.end(...)` lives in the `documentLoadMetricsPromise` `.then()`/`.catch()` handlers inside `_endSpan`, which are only reached through the `load` event chain that `disable()` just severed. The span leaks unexported, and if re-enabled later, `startPageLoadSpan`'s `if (this.pageLoadSpan || ...)` guard would skip creating a new one. Suggest ending `pageLoadSpan` (and clearing `navigationStartTimeMillis`/`pageLoadSpan`) in `disable()` when it's still open.

Human note: Seems like a low-probability edge case (instrumentation disabled specifically during the brief pre-load window). Unsure if its likely that you would call enable() again after disable(). Not a high priority if you feel its not needed.

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
Abinet18: It is very unlikely , but we can consider buffering or some other alternative if it becomes an issue

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
Abinet18: PageLoad span didn't use the active context even before this change, so it should remain a root span

tchoi1131: **Codex review comment**

Thanks. I refreshed and checked the latest PR head (`7d5b4df8`). I see that the intended design is now for `pageLoad` to be a separate root trace, and the current initialization order should produce that result.

One clarification about the previous behavior: the old `pageLoad` call ran synchronously from the overridden `_startSpan(documentLoad)` while upstream `_collectPerformance()` was inside `context.with(propagation.extract(ROOT_CONTEXT, { traceparent }), ...)`. Since `Tracer.startSpan()` defaults an omitted context argument to `api.context.active()`, `pageLoad` inherited a valid document `traceparent`; it was a root only when no valid trace context was present.

Given the new separate-root design, would it be worth making that invariant explicit with `root: true`?

```ts
this.pageLoadSpan = this.tracer.startSpan(PAGE_LOAD_SPAN_NAME, {
	root: true,
	startTime: this.navigationStartTimeMillis,
})
```

A test that starts initialization under an active span and verifies that `pageLoad` has a different trace ID and no parent would also document the intended topology. This is non-blocking, but it would prevent a future initialization-order change from accidentally parenting `pageLoad`.

tchoi1131 (approved):

packages/web/src/instrumentations/splunk-document-load-instrumentation.ts
shumone: We can leave it as is then and return to this if we see its a common issue

shumone (approved):

PR #1895, opened 2026-08-14
[comparison] chore(deps): pnpm-generated picomatch update

mvirgil:
## Purpose

Comparison-only alternative to #1894, showing the final lockfile produced by pnpm rather than a surgical lockfile edit.

## Generation method

1. Temporarily added `picomatch@4.0.3: 4.0.4` as a pnpm override.
2. Ran `pnpm install --lockfile-only --config.fixLockfile=false`.
3. Removed the temporary override.
4. Ran `pnpm install --lockfile-only --config.fixLockfile=false` again.

The final branch has no override or package.json change. Because the disposable worktree could not authenticate to the configured internal registry, generation used the public npm registry.

## Result

- Updates `unplugin@3.0.0` from `picomatch@4.0.3` to `picomatch@4.0.4`.
- Also includes pnpm-generated lockfile normalization and unrelated compatible resolution changes.
- Diff: 6 insertions, 59 deletions.

## Verification

- `pnpm install --lockfile-only --frozen-lockfile`
- `git diff --check`

Do not merge until choosing between this generated diff and the smaller #1894 diff.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1530,105 +1530,89 @@ packages:
     resolution: {integrity: sha512-excjX8DfsIcJ10x1Kzr4RcWe1edC9PquDRRPx3YVCvQv+U5p7Yin2s32ftzikXojb1PIFc/9Mt28/y+iRklkrw==}
     cpu: [arm64]
     os: [linux]
-    libc: [glibc]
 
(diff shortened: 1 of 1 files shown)


mvirgil:
Superseded by #1894, which now uses the scoped picomatch override approach selected after this comparison.

PR #1897, opened 2026-08-14
feat: Blocking instrumentation improvements

shumone:
# Description
This MR fixes 2 gaps in SplunkBlockingElementInstrumentation. 
1. `navigationMetrics.urlOverrides` was ignored. Blocking element selectors and enablement now re-resolve on every SPA route change, so a matching `urlOverrides` entry takes effect after navigation rather than being stuck on the initial page load's configuration. 
2. Open element spans were lost on visibilitychange=hidden (tab switching/etc). Now adds an early visibilitychange handler to interrupt open spans with a completion status of `visibility_hidden`. Existing elements that were closed via that method and are still visible after returning to the page will start a new span. 

Notes: 
- `SplunkBlockingElementInstrumentation` subscribes to a new `NavigationMetricsManager.onRouteChange()` API. Its purpose is to provide a way for attaching to the route change handling logic and re-apply selectors via the callback. 
- Is dependent on the user interactions instrumentation for the spa navigation routechange/history handling. 
- Added a Webkit specific integration test utility (checks on `sendBeacon`); Webkit seems to use `sendBeacon` rather than fetch/xhr when forceflushing spans due to visibilitychange. 

## Type of change

- Documentation update

# How has this been tested?

- Manual testing
- Added unit tests
- Added integration tests


packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
```diff
@@ -86,40 +103,66 @@ export class SplunkBlockingElementInstrumentation extends InstrumentationBase<Sp
...
+	}
+
+	// Re-resolves selectors/enablement and re-applies via watch() — disabled-or-empty is just
+	// "watch nothing," not a distinct branch. Called at enable() and on every route change.
+	private readonly applySelectors = (): void => {
+		this.selectors = isBlockingElementInstrumentationEnabled(this.otelConfig)
+			? resolveBlockingElementSelectors(this.otelConfig)
+			: []
+		this.elementSpanTracker?.setConfiguredSelectors(this.selectors)
+		this.elementVisibilityObserver.watch(this.consumerId, this.selectors, this.handleVisibilityChange)
```
Abinet18, line 142: Codex-Review [P2] Notify replacement selectors before dropping previous selectors

applySelectors() replaces the tracker selector list and then calls watch(), but ElementVisibilityObserver.watch() emits visible:false for removed selectors before scanning newly added selectors. If the same still-visible element matches the old selector and the replacement selector, the removal empties activeSelectorsByElement and ends the span before the replacement is registered. Because the tracker configured list has already changed, that first span can also export browser.element.selector as an empty string, and then a second span opens for the same uninterrupted blocker.

Please add or notify new visible selector matches first, then send disappeared events for dropped selectors. The existing route tests cover dropped, retained, and newly added selectors separately, but not a single physical element transitioning from an old selector to its replacement.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
```diff
@@ -84,21 +83,26 @@ export class ElementSpanTracker {
...
 	}
 
 	/** Total number of currently open (not yet ended) element spans, one per physical element. */
 	get openCount(): number {
 		return this.tracked.size
 	}
 
+	/** Updates the selector list used to order `browser.element.selector` at completion time. */
+	setConfiguredSelectors(selectors: string[]): void {
+		this.configuredSelectors = selectors
```
Abinet18: Codex-Review [P2] Preserve selector history across route configuration changes

setConfiguredSelectors() replaces configuredSelectors for every open span. endSpan() later filters accumulatedSelectors through only this latest list. For an element that matched .spinner and .shared before navigation and remains tracked through .shared after an override, the emitted browser.element.selector drops .spinner even though the README says it contains every selector matched while the span was open.

Please retain a per-span ordered selector history, or equivalent, across configuration changes and add a test that changes configured selectors while an element span remains open.

packages/web/src/instrumentations/splunk-user-interaction-instrumentation.ts
```diff
@@ -204,10 +204,10 @@ export class SplunkUserInteractionInstrumentation extends UserInteractionInstrum
 		span.setAttribute('location.href', newHref)
 		span.setAttribute('prev.href', oldHref)
 
-		if (this.spaMetricsManager) {
+		if (this.navigationMetricsManager) {
 			// Wait for all in-flight resources monitored by SPA metrics to finish loading,
```
Abinet18: Codex-Review [P3] Finish the navigation metrics terminology rename

This comment still says SPA metrics even though the class, field, and public option in this hunk are renamed to navigation metrics. Please update it to avoid leaving stale terminology in the implementation.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
shumone: I think this was just a design choice I made early on, but thinking about it more I'm not sure if its really necessary for it to have to be ordered in the provided config selectors list (and future lists in urlOverrides). It would be simpler to just have it be ordered by when it was added into accumulatedSelectors, which is the current functionality. Let me know what you think and if thats acceptable, i'll just update the readme/comments.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
shumone: Updated by swapping the add/drop loops

packages/web/src/instrumentations/splunk-user-interaction-instrumentation.ts
shumone: Will update comment to use navigation

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
```diff
@@ -86,40 +103,66 @@ export class SplunkBlockingElementInstrumentation extends InstrumentationBase<Sp
...
+
+		this.applySelectors()
 	}
 
 	init(): void {}
 
+	// Called from index.ts on visibilitychange->hidden, before the app's forceFlush() — ending open
+	// spans here (not just on pagehide) so they make it into the export buffer before that flush.
+	// Does not unwatch(): a later visibilitychange->visible reopens still-visible elements.
+	interruptForHidden(): void {
```
Abinet18, line 130: Codex-Review [P2] Do not reopen blocking-element spans while the document is still hidden

interruptForHidden() ends the current spans and clears the instrumentation bookkeeping, but the shared ElementVisibilityObserver remains active. If application code adds or changes a matching element while the tab is still hidden, handleVisibilityChange() can process a new visible:true notification and start another span. That span then includes time spent in the background, and the visible-state resync will not start a fresh span because the element is already tracked.

Please ignore visible:true notifications while document.visibilityState is hidden, then let the existing visibilitychange-to-visible resync start a fresh span. A focused test should interrupt on hidden, mutate or add a matching element before returning to visible, verify no span is open while hidden, and verify tracking begins after the visible resync.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
Abinet18: Ordering selectors by when they are first added to accumulatedSelectors is acceptable and simpler. However, endSpan() currently filters that set through the latest configuredSelectors, so selectors removed by a route override are still lost.
Please serialize the accumulated set directly, document that the order is first-observed/insertion order, and add a route-change test confirming earlier selectors remain on the completed span.

packages/web/src/instrumentations/splunk-blocking-element-instrumentation.ts
shumone: Updated so that those notifications are silently ignored when the visibilityState is hidden. The resync will work as the element that became visible is still being tracked.

packages/web/src/instrumentations/blocking-element/element-span-tracker.ts
shumone: Updated above as well (use accumulatedSelectors list, remove configuredSelectors as its redundant now)

shumone:
Will resolve conflicts after https://github.com/signalfx/splunk-otel-js-web/pull/1888 is merged, as this branch is rebased on top of an older version of that branch.

packages/web/src/index.ts
```diff
@@ -294,11 +294,19 @@ let _deinitSessionTracking: undefined | (() => void)
...
 let _postDocLoadInstrumentation: SplunkPostDocLoadResourceInstrumentation | undefined
 let _navigationMetricsManager: NavigationMetricsManager | undefined
+let _blockingElementInstrumentation: SplunkBlockingElementInstrumentation | undefined
 let eventTarget: InternalEventTarget | undefined
 let _sessionStateUnsubscribe: undefined | (() => void)
 const isLatestTagUsed = isAgentLoadedViaLatestTag()
 const isFullVersionTagUsed = isAgentLoadedViaNextTag() || isAgentLoadedViaLockedVersionTag()
 
+// Registered at module load, before the app-level listener below — wins that race.
+window.addEventListener('visibilitychange', () => {
```
Abinet18: Codex-Review [P2] Guard the module-level visibility listener for non-browser imports
[packages/web/src/index.ts](/Users/adebele/Splunk/splunk-otel-js-web/packages/web/src/index.ts)
This branch adds a top-level window.addEventListener('visibilitychange', ...) before SplunkRum.init() runs. That bypasses the existing non-browser guard in init() at line 430. In SSR/build-time import scenarios, importing the package now throws immediately with ReferenceError: window is not defined, instead of allowing the package to load and letting init() safely no-op/warn.
I confirmed after building:
node -e "require('./packages/web/dist/cjs/index.js')"
# ReferenceError: window is not defined
Suggested fix: wrap the listener registration with a browser-global guard, e.g. typeof window === 'object' && typeof document === 'object', or register it inside init() after the existing window guard while preserving that it is registered before the force-flush visibility listener.

shumone: Added a browser window guard

Abinet18 (approved):

PR #1898, opened 2026-08-21
Feature: Manual page completion api

Abinet18:
# Description

Adds a configuration-free manual page completion API for pages whose readiness cannot be determined accurately from automatic network, media, resource, or loading-element monitoring.
SplunkRum.registerManualPageLoad() returns a navigation-bound handle. Multiple components can register independently, and the page completes after every handle calls markComplete(). The latest accepted completion timestamp determines PCT.
The change also:
Invalidates stale handles after navigation, interruption, timeout, or shutdown.
Supports late registrations during the existing spaMetrics.quietTime window.
Adds the configurable spaMetrics.maxPageLoadTimeoutForManualApi safety timeout, defaulting to 180 seconds.
Records whether completion was manual or automatic.
Finalizes navigation spans before flushing when the page becomes hidden.
Documents API usage, behavior, and limitations.

DEMRUM-6157

## Type of change

- Documentation update

# How has this been tested?

Manual testing using document-load, route-change, multiple-component, late-registration, interruption, timeout, page-hide, and automatic-versus-manual scenarios
Added unit tests for handle lifecycle, timing, timeout, navigation, and PCT relevance behavior
Added integration tests for document-load and route-change completion


packages/web/src/managers/spa-metrics-manager/quiet-period-awaiter.ts
```diff
@@ -207,33 +296,103 @@ export class QuietPeriodAwaiter {
...
+		)
+	}
+
+	private startManualRegistrationTimer(): void {
+		this.clearQuietTimer()
+		this.timeoutId = setTimeout(() => this.resolveManualCompletion(), this.quietTime)
+	}
+
 	private withLoadingResourcesDetails(resolveValue: PageLoadMetricsResolveValue): PageLoadMetricsResult {
+		if (resolveValue.completionSource === 'manual' && this.manualCompletionResourceDetails) {
```
gaurav-v-85: Codex review found 1 medium severity issue:
[M1] Stale resource diagnostics after partial manual completion — [quiet-period-awaiter.ts:392](https://github.com/signalfx/splunk-otel-js-web/pull/1898/files)
markComplete() snapshots resource details for every participant. If participant A completes, participant B remains pending, and the page is interrupted or times out, the manual result reuses A’s old snapshot instead of current loading resources. The emitted span can therefore report stale counts/URLs and hide work active at interruption.
Use the cached snapshot only for completed results, or capture it only when the final participant completes. Add a regression test covering partial completion followed by interruption/timeout.

packages/web/src/managers/spa-metrics-manager/quiet-period-awaiter.ts
gaurav-v-85, line 108: **Suggestion** (non-blocker): Rename `QuietPeriodAwaiter` to `PageLoadCompletionAwaiter`. 

Reason: It has evolved from a quiet-period timer into the page-load completion state machine.
It now owns:
- Automatic quiet-period completion
- Manual participant registration and handle lifecycle
- Manual timeout enforcement
- Page interruption handling
- Completion-source tracking
- Resource snapshots for the final result
- Resolving the shared page-load promise

packages/web/src/managers/navigation-metrics-manager/navigation-metrics-manager.ts
```diff
@@ -393,7 +428,9 @@ export class SpaMetricsManager {
 			getLoadingResourcesCount: () => this.loadingResourcesCount,
 			getLoadingResourceUrls: () => this.loadingResourceUrls,
 			getLongestLoadedResource: () => this.longestLoadedResource,
-			maxPageLoadWaitTime: activeConfig.maxPageLoadWaitTime,
```
gaurav-v-85, line 396: Just want to confirm this was removed intentionally? The automatic load time calculation could be much higher now if I understand this change

packages/web/src/managers/spa-metrics-manager/quiet-period-awaiter.ts
Abinet18: Agree with the naming direction. Since the rename could be broad ,I’d prefer to keep it as a follow-up.

packages/web/src/managers/navigation-metrics-manager/navigation-metrics-manager.ts
Abinet18: Yes, this was intentional. The automatic maxPageLoadWaitTime timer was already disabled on main; this branch only removes the unused pass-through into QuietPeriodAwaiter. Manual API timeout is handled separately through maxPageLoadTimeoutForManualApi. So this should not change automatic load-time timeout behavior.

packages/web/src/managers/spa-metrics-manager/quiet-period-awaiter.ts
Abinet18: handled on latest commit

packages/web/src/managers/navigation-metrics-manager/navigation-metrics-manager.ts
gaurav-v-85: I still see references to maxPageLoadWaitTime in a number of places. If it's dead code, can you file a ticket to properly clean up and deprecate the config? Thanks

gaurav-v-85 (approved):

packages/web/src/managers/navigation-metrics-manager/quiet-period-awaiter.ts
```diff
@@ -163,26 +188,90 @@ export class QuietPeriodAwaiter {
...
 			status: PAGE_LOAD_METRICS_STATUS_INTERRUPTED,
 		})
 	}
 
+	registerManualPageLoad(): ManualPageLoadHandle | undefined {
+		if (this.isResolved) {
+			return undefined
+		}
+
+		if (this.hasManualPageLoadTimedOut()) {
```
shumone, line 224: [P1] Claude: 

Deadline check in `registerManualPageLoad()` can misreport a completed manual page load as timed out.

The deadline check inside `registerManualPageLoad()` always resolves as `TIMEOUT` once past the deadline, without checking whether all participants already completed. The scheduled timeout callback (`startManualPageLoadTimeout()`) does check this — it reports the real completion if `lastManualCompletionTimestamp` is already set and no participants are pending.

This inconsistency means: if a component completes well before the deadline (setting `lastManualCompletionTimestamp` and opening the reopen window), and a new component calls `registerManualPageLoad()` after the deadline elapses but before the scheduled timeout callback has actually run (e.g., a busy main thread or a throttled background tab), that call will discard the real completion and resolve the whole navigation as `TIMEOUT` instead.

Suggest mirroring the same "already completed" check in both places so `registerManualPageLoad()`'s deadline handling matches the scheduled timer's.

packages/web/src/index.ts
```diff
@@ -799,6 +824,14 @@ export const SplunkRum: SplunkOtelWebType = {
 
 	ParentBasedSampler,
 
+	registerManualPageLoad() {
+		if (!inited) {
+			return
```
shumone, line 831: Should this also return a diag.warn (similarly to reportError/getSessionState/etc)? Not a blocker though.

packages/web/src/managers/spa-metrics-manager/spa-metrics-manager.ts
```diff
@@ -233,6 +241,17 @@ export class SpaMetricsManager {
 		}
 	}
 
+	async finalizeCurrentNavigation(endTimestamp = performance.now()): Promise<void> {
+		const pageLoadMetricsPromise = this.pageLoadMetricsPromise
+		this.quietPeriodAwaiter?.interrupt(endTimestamp)
```
shumone: [P2] Claude

In finalizeCurrentNavigation(), quietPeriodAwaiter?.interrupt(endTimestamp) runs outside the try catch block. If it throws synchronously, the async method rejects; its only caller (index.ts's visibilitychange handler) has no .catch() on the .finally() chain, producing an unhandled rejection on the tab-hide/flush path.

packages/web/src/managers/navigation-metrics-manager/navigation-metrics-manager.ts
Abinet18: The config could be reintroduced when plt computation is stable and reasonable, so we can consider removing it if decided we don't need it

packages/web/src/managers/navigation-metrics-manager/quiet-period-awaiter.ts
Abinet18: done

shumone (approved):

PR #1917, opened 2026-09-10
Fix : Avoid reporting resource load errors as error spans, add status to resource spans when available

Abinet18:
# Description

This PR fixes resource load failures being reported as client error spans. Resource failures such as failed image loads can create high-cardinality error IDs because they include resource-specific URLs, while the same failure is better represented on the corresponding resource span.
The change suppresses non-script resource element error events from the error instrumentation and records failed resource status on resource spans instead. Resource spans now receive http.status_code when available, and are marked with span status ERROR when the browser exposes a failing response status. Script-related errors are still reported as client errors to preserve JS syntax/runtime/source-map error visibility.

Fixes DEMRUM-6496.

## Type of change

# How has this been tested?

- Added unit tests
- Added integration tests
- Manual testing


packages/web/src/utils/resource-timing.ts
```diff
@@ -0,0 +1,69 @@
...
+	return undefined
+}
+
+export function isResourceElementLoadError(event: Event): boolean {
+	const target = event.target
+	return (
+		event.type === 'error' &&
+		target instanceof Element &&
+		target.tagName.toUpperCase() !== 'SCRIPT' &&
+		getResourceElementUrl(target) !== undefined
```
shumone, line 55: Codex P1:
This suppresses error events for all recognized resource elements, but the default post-load instrumentation only creates resourceFetch spans for img and script. A dynamically added failing video, audio, or link resource therefore produces neither an error span nor a resource span. Retain the client-error span unless the active resource instrumentation can emit the replacement span, and add coverage for a default-config dynamic non-image resource failure.

Abinet18: If a resource type is not enabled for resource monitoring, reporting its load failure as a generic client error is also misleading. These element load failures usually have no stack trace and are better represented by resource spans when that resource type is monitored. For unmonitored resource types, I’d prefer not to preserve them as error spans because that reintroduces the high-cardinality/noisy behavior this PR is trying to avoid.

shumone (approved):

PR #1921, opened 2026-09-14
fix: add loaf entry start time to loaf span

Abinet18:
# Description

Adds LoAF entry start timing to Long Animation Frame spans so downstream consumers can place attributed scripts relative to the LoAF span start.

The agent now exports `loaf.entry_start_time` once per LoAF span. Existing `loaf.script[n].start_time` values are preserved. Both values use the browser performance time origin, so consumers can calculate script timing as:

```text
scriptOffsetFromLoafStart =
  loaf.script[n].start_time - loaf.entry_start_time

scriptAbsoluteStart =
  loafSpan.start + loaf.script[n].start_time - loaf.entry_start_time
```

Fixes DEMRUM-6205

## Type of change

## How has this been tested?
Tests added/updated:
- Updated LoAF span attribute unit coverage to verify loaf.entry_start_time is emitted and rounded correctly.
- Updated LoAF span attribute unit coverage to keep normal script fixtures within the parent LoAF timing window.
- Updated Long Animation Frame browser integration coverage to verify emitted LoAF spans include loaf.entry_start_time and script timing values needed to calculate script position relative to the LoAF span.


packages/web/src/instrumentations/loaf/span-attributes.ts
```diff
@@ -48,6 +48,7 @@ export function setLoafEntryAttributes(span: Span, entry: PerformanceLongAnimati
 		setRoundedNumberAttribute(span, `${prefix}.source_char_position`, script.sourceCharPosition)
 		setStringAttribute(span, `${prefix}.source_url`, script.sourceURL)
 		setStringAttribute(span, `${prefix}.source_function_name`, script.sourceFunctionName)
+		setRoundedNumberAttribute(span, `${prefix}.offset`, script.startTime - entry.startTime)
```
shumone, line 51: non-blocking: 

I don't think it's possible for the script startTime to be earlier than the entry.startTime, but maybe a non-negative check would be useful? If you don't think its necessary or possible at all then can ignore this comment.

Abinet18: Good point. For valid LoAF entries this should not happen because both timestamps are from the same performance timeline and the script is part of the entry, but I added a defensive non-negative clamp for the new offset. I kept start_time unchanged and only clamp finite offsets, so malformed non-finite timing still gets omitted by the existing numeric attribute handling.

shumone (approved):

packages/web/src/instrumentations/loaf/span-attributes.ts
```diff
@@ -48,6 +49,7 @@ export function setLoafEntryAttributes(span: Span, entry: PerformanceLongAnimati
 		setRoundedNumberAttribute(span, `${prefix}.source_char_position`, script.sourceCharPosition)
 		setStringAttribute(span, `${prefix}.source_url`, script.sourceURL)
 		setStringAttribute(span, `${prefix}.source_function_name`, script.sourceFunctionName)
+		setRoundedNumberAttribute(span, `${prefix}.offset`, Number.isFinite(offset) ? Math.max(0, offset) : offset)
```
tchoi1131: **Codex-generated review comment:** If we keep the per-script offset, could we omit the attribute when the calculated offset is negative rather than clamp it to zero?

A negative value indicates invalid or inconsistent timing. Changing it to zero creates valid-looking evidence that says the script started exactly at the LoAF span start, potentially causing incorrect downstream placement.

packages/web/src/instrumentations/loaf/span-attributes.test.ts
```diff
@@ -66,6 +66,7 @@ describe('LoAF span attributes', () => {
 					createScript({ duration: 20, sourceURL: 'blob:https://example.com/id?kept=true#kept' }),
 					createScript({ duration: 30, sourceURL: '<anonymous>' }),
 				],
+				startTime: 150,
```
tchoi1131, line 74: **Codex-generated review comment:** Could we update this fixture to represent valid LoAF timing?

The script starts `10.11 ms` after the LoAF begins and lasts `161.3 ms`, while the LoAF duration is only `161.38 ms`. This places the script end outside the LoAF. Other scripts use the default `startTime = 0`, which is earlier than the entry start and exercises the clamp unintentionally.

The normal test should satisfy:

```text
script.startTime >= entry.startTime
script.startTime - entry.startTime + script.duration <= entry.duration
```

Malformed timing can remain covered by a separate defensive test.

packages/integration-tests/src/tests/loaf/loaf.spec.ts
```diff
@@ -59,6 +59,7 @@ test.describe('long animation frame', () => {
 			expect(loafSpan).toHaveNumericAttribute(`loaf.script[${index}].forced_style_and_layout_duration`)
 			expect(loafSpan).toHaveStringAttribute(`loaf.script[${index}].invoker`)
 			expect(loafSpan).toHaveStringAttribute(`loaf.script[${index}].invoker_type`)
+			expect(loafSpan).toHaveNumericAttribute(`loaf.script[${index}].offset`)
```
tchoi1131: **Codex-generated review comment:** Could this test verify the meaning of the new value rather than only its presence? For every exported script, we could assert with a small browser-timing tolerance:

```text
offset >= 0
offset <= loaf.duration
offset + script.duration <= loaf.duration + tolerance
```

Otherwise, this test would still pass if every offset were incorrectly reported as zero.

packages/web/README.md
```diff
@@ -236,7 +236,7 @@ SplunkRum.init({
...
+LoAF spans include frame timing attributes such as `loaf.duration`, `loaf.blocking_duration`, `loaf.paint_time`, `loaf.presentation_time`, `loaf.render_start`, `loaf.style_and_layout_start`, and `loaf.first_ui_event_timestamp`. They also include bounded script attribution: `loaf.script_count` reports the original number of scripts, while only the top three scripts by duration are exported as `loaf.script[0..2].*` attributes, including script timing fields such as `duration`, `start_time`, `offset`, `execution_start`, `pause_duration`, `forced_style_and_layout_duration`, and `source_char_position`. `start_time` preserves the browser's performance-origin-relative script timestamp, while `offset` reports the script start relative to the LoAF span start. Script source attribution is reported as provided by the browser; use `exporter.onAttributesSerializing` if your application needs to redact or transform those attributes before export.
```
tchoi1131: **Codex-generated review comment:** Please document that this timing value is measured in milliseconds, its reference point, and the reconstruction formula. Because this becomes a backend-facing telemetry contract, making the clock basis explicit will prevent consumers from treating it as an absolute timestamp.

packages/web/src/instrumentations/loaf/span-attributes.ts
```diff
@@ -40,6 +40,7 @@ export function setLoafEntryAttributes(span: Span, entry: PerformanceLongAnimati
 
 	getLoafScriptSummaries(scripts).forEach((script, index) => {
 		const prefix = `loaf.script[${index}]`
+		const offset = script.startTime - entry.startTime
```
tchoi1131: Could we export the LoAF entry’s `startTime` once as a span-level attribute, such as `loaf.entry_start_time`, instead of adding `loaf.script[n].offset` for every exported script?

The LoAF entry and its scripts use the same performance time origin. Since `loaf.script[n].start_time` is already exported, the backend could calculate:

```text
scriptAbsoluteStart =
    loafSpan.start
    + loaf.script[n].start_time
    - loaf.entry_start_time
```

This would add one attribute per LoAF span instead of up to three derived attributes. Is there any reason to add the offset for each script?

Abinet18: Good suggestion. I updated the implementation to export loaf.entry_start_time once at the LoAF span level instead of exporting loaf.script[n].offset for each script.

packages/web/src/instrumentations/loaf/span-attributes.ts
Abinet18: This is no longer applicable after the update.

packages/web/src/instrumentations/loaf/span-attributes.test.ts
Abinet18: Updated. The normal fixture now uses valid LoAF timing so the top script starts after the LoAF entry start and ends within the LoAF duration. Malformed timing remains covered separately by the defensive test.

packages/integration-tests/src/tests/loaf/loaf.spec.ts
Abinet18: Updated

packages/web/README.md
Abinet18: Updated the README to document that loaf.entry_start_time and loaf.script[n].start_time are both milliseconds from the browser performance time origin. It also includes formulas for calculating script start relative to the LoAF span and reconstructing the absolute script start.

packages/web/src/instrumentations/loaf/span-attributes.test.ts
```diff
@@ -66,12 +67,14 @@ describe('LoAF span attributes', () => {
 					createScript({ duration: 20, sourceURL: 'blob:https://example.com/id?kept=true#kept' }),
```
tchoi1131: **Codex-generated follow-up review comment:** This normal fixture still includes two exported scripts with the default `startTime = 0` while the entry uses `startTime = 150`. That means `script[1]` and `script[2]` start before their parent LoAF. Could we assign valid `startTime` values to every script in this normal fixture and keep malformed timing only in the dedicated defensive test?

packages/web/README.md
```diff
@@ -236,7 +236,7 @@ SplunkRum.init({
...
+LoAF spans include frame timing attributes such as `loaf.duration`, `loaf.blocking_duration`, `loaf.entry_start_time`, `loaf.paint_time`, `loaf.presentation_time`, `loaf.render_start`, `loaf.style_and_layout_start`, and `loaf.first_ui_event_timestamp`. They also include bounded script attribution: `loaf.script_count` reports the original number of scripts, while only the top three scripts by duration are exported as `loaf.script[0..2].*` attributes, including script timing fields such as `duration`, `start_time`, `execution_start`, `pause_duration`, `forced_style_and_layout_duration`, and `source_char_position`. `loaf.entry_start_time` and `loaf.script[n].start_time` are both measured in milliseconds from the browser performance time origin. Consumers can calculate the script start relative to the LoAF span start as `loaf.script[n].start_time - loaf.entry_start_time`, or reconstruct the absolute script start as `loafSpan.start + loaf.script[n].start_time - loaf.entry_start_time`. Script source attribution is reported as provided by the browser; use `exporter.onAttributesSerializing` if your application needs to redact or transform those attributes before export.
```
tchoi1131: **Codex-generated follow-up review comment:** Could we state that `loafSpan.start` must be converted to milliseconds before applying this formula? The two LoAF attributes are milliseconds from the browser performance time origin, while an exported OpenTelemetry span start may use a different representation or unit, such as Unix nanoseconds. Without that normalization, the formula is dimensionally ambiguous.

tchoi1131 (approved):
Approved. The inline findings are low-severity, non-blocking follow-ups.

**Codex-generated follow-up review comment:** The “How has this been tested?” section still says `loaf.script[n].offset` is emitted and tested. The implementation now exports `loaf.entry_start_time` instead, so could you update those bullets to match the final telemetry contract?

packages/web/README.md
Abinet18: done

packages/web/src/instrumentations/loaf/span-attributes.test.ts
Abinet18: updated

PR #1922, opened 2026-09-16
fix: memory leak in instrumentation-user-interaction

belzile:
# Description

Non Claude: fixes a memory leak by turning a Map into a WeakMap. The Map prevents garbage collection of some detached elements. The fix has already been implemented in the copied over dependency (see references below)

Claude:

`UserInteractionInstrumentation` tracks the listeners it wraps in `_wrappedListeners`, a
`WeakMap<listener, Map<type, Map<Element, wrapper>>>`. The outer level is weak, but the innermost
level — element to wrapper — is a strong `Map`, so every element a listener has ever been added to
is retained by that map until `removeEventListener` is called for that exact element and type.

Whenever a listener outlives the elements it was attached to, that is a leak. It is the normal case
for a framework that registers a method taken from a shared prototype: the function is immortal, so
the map under it only ever grows. `@microsoft/fast-foundation` does this in
`FormAssociated.connectedCallback`, which adds the prototype's `keypress` handler and never removes
it — one retained entry per form-associated element ever created, each one holding its element and
the detached subtree hanging off it. This is not limited to instrumented events:
`addPatchedListener` runs before the `eventNames` check, which only gates span creation, so an event
type nobody asked to instrument is still tracked.

The fix makes the element level a `WeakMap`, so an element that is dropped from the DOM can be
collected while the listener is still alive. Two consequences follow:

- A `WeakMap` has no `size`, so the cascade that pruned empty per-type and per-listener entries is
  gone. What remains is one empty `Map` entry per event type a given listener was used with, and the
  outer `WeakMap` still drops the whole chain when the listener itself is collected.
- The element is now a `WeakMap` key, which a receiver that is not an object cannot be. When a
  caller detaches `addEventListener` from its element and calls it bare, the receiver is `undefined`,
  which the browser resolves to the global object rather than rejecting — so the call must be
  forwarded to the original method uninstrumented instead of used as a key.

Both mirror the equivalent upstream changes in opentelemetry-js-contrib, which this directory is
vendored from: open-telemetry/opentelemetry-js-contrib#3576 for the weak bookkeeping and
open-telemetry/opentelemetry-js-contrib#3653 for the receiver guard, released upstream in 0.65.0 and
0.66.0 respectively. The vendored copy predates both.

No Github issue to link; this was found while investigating a growing heap in an application built
on FAST web components.

## Type of change

## How has this been tested?

- Added unit tests

`packages/web/tests/user-interaction-instrumentation.test.ts` covers the bookkeeping directly: that
the element level is weak, that a detached `addEventListener` still registers and removes, and five
regression guards for the behaviour that had to survive the change (a shared listener patched on
every element it is added to, removal affecting only the element it was removed from, re-patching
after removal, a single registration per element/type/listener, and a `once` listener cleared from
the bookkeeping after it fires). Both new assertions were confirmed to fail against the unmodified
source, in all three browsers.

Verified with `pnpm run lint`, the root typecheck, the full unit suite (1926 passing across
chromium, firefox and webkit) and the Playwright integration suite on chromium (171 passing,
including all 29 `user-interaction` specs).

## AI usage

Check one option.

- [x] AI was used to create the full change

## Checklist

- [x] The selected test coverage matches the risk and browser impact of the change — the patched code
      path is shared by every event listener on the page, and the new tests run in chromium, firefox
      and webkit alongside the existing integration specs.
- [x] Documentation, changelog, or release notes have been updated when customer-facing behavior
      changes — no consumer-visible behaviour changes, and changelog entries are added per release
      with a link to the PR.
- [x] Backward compatibility, privacy, telemetry volume, and cardinality impact have been considered —
      internal bookkeeping only. The public API, the spans produced, their attributes and their volume
      are unchanged; no new data is read from the page.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


belzile:

Abinet18:
@belzile Thanks for submitting this change, I just rebased it and pushed it with my signature , will ask for internal review

Abinet18:
> I have read the CLA Document and I hereby sign the CLA
> 
> Also check with your team whether Ticketmaster has already signed a corporate CLA with Splunk for work-time contributions — signing an individual CLA for code written on company time is the part worth a human decision, not mine. I haven't touched your git config, history, or the PR.

I checked with Splunk’s OSS team. Splunk generally does not use corporate CLAs for its OSS repositories and relies on individual CLAs instead, so your signed individual CLA is sufficient from Splunk’s side. They cannot confirm Ticketmaster’s internal policy, so you may still want to verify whether Ticketmaster requires separate approval for work-time contributions.

belzile:
> > I have read the CLA Document and I hereby sign the CLA
> > Also check with your team whether Ticketmaster has already signed a corporate CLA with Splunk for work-time contributions — signing an individual CLA for code written on company time is the part worth a human decision, not mine. I haven't touched your git config, history, or the PR.
> 
> I checked with Splunk’s OSS team. Splunk generally does not use corporate CLAs for its OSS repositories and relies on individual CLAs instead, so your signed individual CLA is sufficient from Splunk’s side. They cannot confirm Ticketmaster’s internal policy, so you may still want to verify whether Ticketmaster requires separate approval for work-time contributions.

To confirm because it does not looks clear from my interpretation of your message: do you need me to do something in relation to this MR to get it merged?

Abinet18:
> > > I have read the CLA Document and I hereby sign the CLA
> > > Also check with your team whether Ticketmaster has already signed a corporate CLA with Splunk for work-time contributions — signing an individual CLA for code written on company time is the part worth a human decision, not mine. I haven't touched your git config, history, or the PR.
> > 
> > 
> > I checked with Splunk’s OSS team. Splunk generally does not use corporate CLAs for its OSS repositories and relies on individual CLAs instead, so your signed individual CLA is sufficient from Splunk’s side. They cannot confirm Ticketmaster’s internal policy, so you may still want to verify whether Ticketmaster requires separate approval for work-time contributions.
> 
> To confirm because it does not looks clear from my interpretation of your message: do you need me to do something in relation to this MR to get it merged?

If you are ok contributing this change as an individual with individual CLA , there is nothing needed. But if you want your contribution as an org, we don't support the corresponding corporate CLA.

belzile:
Ok, yes I am OK contributing with my individual CLA.

belzile:
I just realized I copied the whole paragraph not meant to be copied from the Claude output on the CLA comment after I asked it what to do after the MR turned red for that CLA error thing...

